PR #1, opened 2016-03-13
Update Service structs to match API responses

paulspringett:
This PR has a few changes to the `GetService` func and the `Service` and `Version` structs:

* Introduces a new `ServiceDetail` struct as the response from `/service/:id/detail` differs from other service endpoints meaning the `Service` struct cannot be used. Specifically in one the `version` is a number and in the other a version object.
* Changes the `Service` struct tag for the `ActiveVersion` attribute from `active_version` to `version`. The Fastly docs [incorrectly document the field name](https://community.fastly.com/t/documentation-for-get-service-appears-incorrect/594)
* Removes the timestamp attributes from the `Service` struct as the API responses do not include them
* Adds a `Versions` slice to both `Service` and `ServiceDetail` structs listing all versions
* Remove the timestamp attributes from the `Version` struct as they weren't being parsed correctly

Let me know if you want to discuss these in more detail or split them up into separate PRs, thanks :)

### DIFF
--- a/service.go
+++ b/service.go
@@ -3,7 +3,6 @@ package fastly
 import (
 	"fmt"
 	"sort"
-	"time"
 )
 
 // Service represents a single service for the Fastly account.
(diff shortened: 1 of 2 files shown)


sethvargo:
Hey @paulspringett 

I found a few inconsistencies between the fastly docs and the actual API responses. If you email support@, they are pretty good about fixing them though :smile:

PR #2, opened 2016-03-13
Return error message on validate failure

paulspringett:
If the version validate call fails, include the error message from Fastly in the return arguments.


sethvargo:
Hey @paulspringett 

Thanks! This looks good. Test failures look like API rate limits :frowning:

PR #5, opened 2016-04-01
Additions to Service, Domain, and Backend

catsby:
- Service now supports `deleted_at`; added deleted, created, updated at
- Updates documentation on domain; it's for the Service, not account
- Added Hostname to backend; if a hostname is used in address, it will show up
  here. If an IP is used, hostname will be empty

I had difficulty running tests... it seems they are hardcoded to an account? 
I took a spike at refactoring that, but I don't know a way to clean up a service after each test/suite


sethvargo:
Hey @catsby,

They are hardcoded to an account. I'm going to try to setup something like VCR to record and commit the responses soonish

catsby:
Thanks Seth!

PR #7, opened 2016-04-20
Gzips: Allow empty ContentTypes, Extensions

catsby:
This PR removes the `omitempty` attribute of `ContentTypes` and `Extensions` for the `Gzip` struct. In developing the Fastly Provider for Terraform, I've found that _not_ submitting a value for either of these will result in that attribute receiving a default configuration:

``` 
# omit content_types
$ curl -s -H "Fastly-Key: ******" -X POST -d "name=thing&extensions=css" https://api.fastly.com/service/60wRp8bq9WXSPwANtoWaFx/version/1/gzip | jq . 
{ 
"name": "thing", 
"extensions": "css", 
"service_id": "60wRp8bq9WXSPwANtoWaFx", 
"version": "1", 
"content_types": "text/html application/x-javascript text/css application/javascript text/javascript application/json application/vnd.ms-fontobject application/x-font-opentype application/x-font-truetype application/x-font-ttf application/xml font/eot font/opentype font/otf image/svg+xml image/vnd.microsoft.icon text/plain text/xml", 
"cache_condition": "" 
}

# omit extensions
$ curl -s -H "Fastly-Key: ******" -X POST -d "name=otherthing&content_types=text/html" https://api.fastly.com/service/60wRp8bq9WXSPwANtoWaFx/version/1/gzip | jq . 
{ 
"name": "otherthing", 
"content_types": "text/html", 
"service_id": "60wRp8bq9WXSPwANtoWaFx", 
"version": "1", 
"cache_condition": "", 
"extensions": "css js html eot ico otf ttf json" 
} 
$ 
```

I opened a support ticket with Fastly and got work that this was intentional, and the solution is to send an empty string... I can share the convo privately if you'd like, but here's the reply:

fastlysupport

I've added a test to confirm the omissions, but I'm having trouble running the tests. I also fix what I think is a typo in the tests; in the `Get` call we create a new `ngzip` struct to receive the value, but we do the checks against prior `gzip` struct. The gzip fixtures need to be recreated.

### DIFF
--- a/gzip.go
+++ b/gzip.go
@@ -67,8 +67,8 @@ type CreateGzipInput struct {
 	Version string
 
 	Name           string `form:"name,omitempty"`
-	ContentTypes   string `form:"content_types,omitempty"`
-	Extensions     string `form:"extensions,omitempty"`
+	ContentTypes   string `form:"content_types"`
+	Extensions     string `form:"extensions"`
 	CacheCondition string `form:"cache_condition,omitempty"`
 }
 
--- a/gzip_test.go
+++ b/gzip_test.go
@@ -80,13 +80,130 @@ func TestClient_Gzips(t *testing.T) {
 	if err != nil {
 		t.Fatal(err)
 	}
-	if gzip.Name != ngzip.Name {
+	if ngzip.Name != ngzip.Name {
+		t.Errorf("bad name: %q", ngzip.Name)
+	}
+	if ngzip.ContentTypes != "text/html text/css" {
+		t.Errorf("bad content_types: %q", ngzip.ContentTypes)
+	}
+	if ngzip.Extensions != "html css" {
+		t.Errorf("bad extensions: %q", ngzip.Extensions)
+	}
+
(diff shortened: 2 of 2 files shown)


sethvargo:
Hey @catsby 

I took your commit and opened up #8. Working with the fixtures is annoying because you need a new API key each time, and there's no way to add you as a "collaborator" in fastly.

The test is actually correct as written, but it's confusing. What I'm doing is asserting that the response from the GET is the same as the response from the CREATE (so comparing ngzip's values to gzip's values). I switched the ordering to try and make that clearer. 

Unfortunately the fixture system requires that we re-create all the gzip fixtures, so there's a bit of noise in the other PR, but all the tests pass. Let me know if that looks good to you! 😄

catsby:
> The test is actually correct as written, but it's confusing

Ah, I saw ` gzip.ContentTypes != "text/html text/css" ` after we did `GET` into `nzip` , thus the change.

PR #9, opened 2016-04-25
Add Domain attribute to S3 logging struct

catsby:
It's explained in the user guide, but missing from the API docs:

- Guide: https://docs.fastly.com/guides/streaming-logs/log-streaming-amazon-s3
- API Docs: https://docs.fastly.com/api/logging#logging_s3

I have a support issue opened to confirm. I've confirmed the field exists and is
usable via curl, just waiting to hear if the omission is for a reason or just a
mistake.

### DIFF
--- a/s3.go
+++ b/s3.go
@@ -13,6 +13,7 @@ type S3 struct {
 
 	Name              string     `mapstructure:"name"`
 	BucketName        string     `mapstructure:"bucket_name"`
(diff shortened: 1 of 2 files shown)


sethvargo:
@catsby wonderful lol. Ping me when you get a result.

catsby:
Confirmed it's just an omission; they will update the docs

Issue #11, opened 2016-06-27
client.RequestForm clobbers intentional uses of '\.'


alienth:
This function attempts to workaround a quirk of the form library where cases of `.` characters are replaced with `%5C.`. 

https://github.com/sethvargo/go-fastly/blob/master/client.go#L164

Problem is when it does this, it replaces intentional cases of `\.`. For example, a VCL I'm uploading has a `\.` within a regex, which this code clobbers and replaces with a bare `.`.

sethvargo:
Hi @alienth 

Can you please submit a test that shows the failling behavior and I'll fix it up right away?

alienth:
@sethvargo , here ya go. Thanks for looking into this!

```
package main

import (
        "fmt"
        "github.com/sethvargo/go-fastly"
)

func main() {
        client, _ := fastly.NewClient("foo")
        myvcl := "if (req.http.host ~ \"([[:alnum:]]+)\\.([[:alnum:]]+)$\") {"
        fmt.Println(myvcl)
        vcl, _ := client.UpdateVCL(&fastly.UpdateVCLInput{Name: "myvcl", Service: "abcdef0123456789", Version: "1", Content: myvcl})
        fmt.Println(vcl.Content)
}
```

The output of this will be as follows:

```
if (req.http.host ~ "([[:alnum:]]+)\.([[:alnum:]]+)$") {
if (req.http.host ~ "([[:alnum:]]+).([[:alnum:]]+)$") {
```

You can also verify on Fastly that the resulting VCL is missing the all important `\`.

I've verified that ripping the `strings.Replace` out of `RequestForm()` fixes this, and it also doesn't appear to cause any unwanted escapes. You can see an example of the form lib DTRT in my comment about.

sethvargo:
Hi @alienth 

Sadly this is not fixed upstream. Take a look at `settings.go`. The form keys use a "." such as "general.default_ttl". When removing that value, the submitted form is:

```
Service=xxx&Version=1&general%5C.default_ttl=1234
```

Fastly expects that to be a dot, and the settings fail 😦

alienth:
Ahh, so perhaps this only happens for form keys? As you saw in my test, values seemed to be AOK.

sethvargo:
Maybe, but I'm still not really following the issue. Also, there's no way to just reverse the behavior on just the keys.

I don't think that function is stripping out the slash... it shouldn't touch slashes at all.

alienth:
[...]

PR #12, opened 2016-06-27
Only undo '.' escaping in field names.

alienth:
Fixes cases where content which is intended to be escaped, such as VCLs containing `\.`, becomes clobbered, as described in #11.

### DIFF
--- a/client.go
+++ b/client.go
@@ -158,10 +158,17 @@ func (c *Client) RequestForm(verb, p string, i interface{}, ro *RequestOptions) 
 	}
 	ro.Headers["Content-Type"] = "application/x-www-form-urlencoded"
 
-	// There is a super-jank implementation in the form library where fields with
-	// a "dot" are replaced with "/.". That is then URL encoded and Fastly just
-	// dies. We fix that here.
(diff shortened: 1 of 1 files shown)


alienth:
For the following test:

```
package main

import (
        "fmt"
        "github.com/sethvargo/go-fastly"
)

func main() {
        client, _ := fastly.NewClient("foo")
        myvcl := "if (req.http.host ~ \"([[:alnum:]]+)\\.([[:alnum:]]+)$\") {"
        fmt.Println(myvcl)
        vcl, _ := client.UpdateVCL(&fastly.UpdateVCLInput{Name: "myvcl", Service: "abcdef0123456789", Version: "1", Content: myvcl})
        fmt.Println(vcl.Content)
}
```

Current output:
```
if (req.http.host ~ "([[:alnum:]]+)\.([[:alnum:]]+)$") {
if (req.http.host ~ "([[:alnum:]]+).([[:alnum:]]+)$") {
```

Output with changes in PR:

```
if (req.http.host ~ "([[:alnum:]]+)\.([[:alnum:]]+)$") {
if (req.http.host ~ "([[:alnum:]]+)\.([[:alnum:]]+)$") {
```

PR #16, opened 2016-06-29
Update form dep and remove workaround.

alienth:
The author of form added a function which allows us to specify a custom delimiter for its internal use. This PR will cause go-fastly to make use of that and set the delimiter to `|`, which is not used in any field keys for the Fastly API.


sethvargo:
bdcb612

PR #17, opened 2016-07-02
Use Compatibools for IgnoreIfSet input.

alienth:
Right now it is impossible to set this field to true up in Fastly since it is sending up 'true' strings, which fastly will silently translate to '0'. This change fixes that.

Note that if anyone is using this lib is setting this field to `true`, this change will make that actually happen, which may result in unexpected consequences.


sethvargo:
Ping @catsby - this okay?

catsby:
It is! Thanks @alienth

PR #21, opened 2016-07-27
Close the resp body to prevent socket leaks.

alienth:
Without this, long-running programs will build up dead sockets, eventually leading to FD limits.

### DIFF
--- a/backend.go
+++ b/backend.go
@@ -67,6 +67,7 @@ func (c *Client) ListBackends(i *ListBackendsInput) ([]*Backend, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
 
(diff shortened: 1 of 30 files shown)


sethvargo:
Hi there,

The `[decodeJSON](https://github.com/sethvargo/go-fastly/blob/master/client.go#L189)` function closes the response body.

alienth:
Oh! Well, something is amiss then. I have a long-running app which filled up with dead sockets left by connections to Fastly.

Will dig further.

sethvargo:
@alienth did closing the response body like this resolve the issue?

alienth:
I thought it did, but after waiting a bit longer I saw more sockets pile up. Opened a separate issue to troubleshoot.

PR #26, opened 2016-10-26
Use pointer to bool on backend requests

zimbatm:
Because the "empty" value is false, it's not possible to send that value
to Fastly. And Fastly defaults that value to be true some times (eg:
ssl_check_cert).


zimbatm:
I didn't understand how the form encoding works. #27 actually fixes the issue.

Issue #32, opened 2017-02-08
[BUG] Cannot clear certain cache keys


kf6nux:
[This line](https://github.com/sethvargo/go-fastly/blob/master/request.go#L34) alters keys that begin with a slash or contain more than one slash in a row.

E.g.1: A call to
`/service/yourservice/purge//keyStartingWithASlash`
becomes
`/service/yourservice/purge/keyStartingWithASlash`

E.g.2: A call to
`/service/yourservice/purge/keyWithDouble//Slash`
becomes
`/service/yourservice/purge/keyWithDouble/Slash`

It appears the intended behavior at this line is something more along the lines of "add or remove slashes as needed at the joined location when merging the host path with the api path" or in code: 57dedafea40219991de4cdf582222ffbf948185d

Tangentially related:
Due to #14, it's worth noting that Fastly [explicitly states](https://community.fastly.com/t/should-i-url-encode-the-purge-key-when-using-the-api/301/2?u=benjamin_zarzycki) that keys should not be URL encoded.  If/when a fix for #14 does URL encode data parameters to the Fastly API, that fix must not URL encode cache keys.

kf6nux:
Thanks for the merge @sethvargo !  I'll close this issue out.

Issue #14, opened 2016-06-29
URL components are not urlencoded, resulting in API failures for objects with certain names


alienth:
I have a cache condition with a name containing a character which needs to be urlencoded. (ex: a `text/html` condition). When I execute the following code which simply lists and tries to delete conditions which are found, it will fail:

```
        existingCacheSettings, err := client.ListCacheSettings(&fastly.ListCacheSettingsInput{Service: "abcd", Version: 10})
        for _, setting := range existingCacheSettings {
                err := client.DeleteCacheSetting(&fastly.DeleteCacheSettingInput{Service: "abcd", Name: setting.Name, Version: 10})
                if err != nil {
                        return err
                }
        }
```

The reason it is failing is because it ends up calling the following URL:

```
/service/abcd/version/10/condition/text/html
```

(note the unescaped forward-slash)

Any parameters which go into forming an API call URL should be urlescaped to prevent issues such as this. These object names are perfectly valid from Fastly's POV, but will cause go-fastly to malfunction.

I tried to address this in go-fastly by wrapping all URL components which could contain these characters in an escape method from `net/url`. Unfortunately, name components in the Fastly API may contain not only forward slashes but also whitespace, and `net/url` has no method to deal with both possible cases. Fixing this in `go-fastly` may require that we do a bit of a hacky `strings.Replace()`.

alienth:
Bit of a stretch, but I've submitted a case to `net/url` to see if this type of escaping functionality can be supported: https://github.com/golang/go/issues/16207

alienth:
For now I think the best option avail is to write a small function which calls `url.Parse()` on URL components, and also encodes the forward-slashes. Started work on it this evening. I'll plan on sending a PR tomorrow.

sethvargo:
Looks like go 1.8 might fix this: https://github.com/golang/go/commit/7e2bf952a905f16a17099970392ea17545cdd193

msuterski:
Did this ever got fixed (in go 1.8)?

I run into similar problem in Terraform 0.9.4 with conditions that included `/` in names.

PR #33, opened 2017-02-08
Resolves Issue #32

kf6nux:
See issue #32 for details.

### DIFF
--- a/request.go
+++ b/request.go
@@ -4,7 +4,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"path"
+	"strings"
 )
 
 // RequestOptions is the list of options to pass to the request.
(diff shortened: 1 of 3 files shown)


request.go
```diff
@@ -31,7 +31,7 @@ func (c *Client) RawRequest(verb, p string, ro *RequestOptions) (*http.Request,
 
 	// Append the path to the URL.
 	u := *c.url
-	u.Path = path.Join(c.url.Path, p)
+	u.Path = strings.TrimRight(c.url.Path, "/") + "/" + strings.TrimLeft(p, "/")
```
sethvargo: As far as I can tell, this preserves the current behavior. If I read your issue correctly, you don't want to alter `p` at all, so I think this needs to become:

```golang
u.Path = strings.TrimRight(c.url.Path, "/") + "/" + p
```

sethvargo:
It would also be great to get a test for this change, to prevent a regression in the case of a fix for #14 in the future.

request.go
kf6nux: This preserves the current behavior _only at the join location_.  `p` remains unchanged.

kf6nux:
Tests have been added and the existing Purge test has been fixed.

PR #38, opened 2017-03-22
Support alternate endpoints

whiteley:
Happy with any other solution that achieves the same thing but thought I'd at least take a crack at it vs. lobbing an issue.

### DIFF
--- a/client.go
+++ b/client.go
@@ -64,20 +64,24 @@ func DefaultClient() *Client {
 	return client
 }
 
-// NewClient creates a new API client with the given key. Because Fastly allows
-// some requests without an API key, this function will not error if the API
-// token is not supplied. Attempts to make a request that requires an API key
-// will return a 403 response.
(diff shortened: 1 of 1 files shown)


sethvargo:
Hey @whiteley 

Thank you for the PR. I think `Address` is a public field on the client. Is there a reason you can't do this:

```go
client, err := NewClient("abcd1234")
if err != nil {
  panic(err)
}
client.Address = "my.custom.endpoint"
```

whiteley:
I'll have to go back and double check but iirc the init function set url from Address and setting it after wouldn't work and you couldn't call init again since it's private.

Sorry if I'm misremembering or made a mistake in looking at it.

whiteley:
@sethvargo I'm far from a Go expert but I had a more experienced friend check what I was doing and they agreed it seems impossible at the moment to address another API server. Could you take another look at this? My end goal is to add the support to the Fastly provider in Terraform. Thanks!

Issue #36, opened 2017-02-23
Bug: LatestVersion


Integralist:
LatestVersion for me is reporting as `99` when in fact it's `401`.

I think this is because when I print out the listed versions I see:

```
&{Number:4 Comment: ServiceID:123 Active:false Locked:true Deployed:false Staging:false Testing:false}
&{Number:40 Comment: ServiceID:123 Active:false Locked:true Deployed:false Staging:false Testing:false}
&{Number:400 Comment: ServiceID:123 Active:false Locked:true Deployed:false Staging:false Testing:false}
&{Number:401 Comment: ServiceID:123 Active:true Locked:true Deployed:false Staging:false Testing:false}
&{Number:41 Comment: ServiceID:123 Active:false Locked:true Deployed:false Staging:false Testing:false}
```

The last few iterations are:

```
&{Number:9 Comment: ServiceID:123 Active:false Locked:true Deployed:false Staging:false Testing:false}
&{Number:90 Comment: ServiceID:123 Active:false Locked:true Deployed:false Staging:false Testing:false}
...others...
&{Number:99 Comment: ServiceID:123 Active:false Locked:true Deployed:false Staging:false Testing:false}
```

So looks like the Number isn't sorted correctly. Maybe the `latestVersion` call isn't sorting correctly?

sethvargo:
Hmmm... I think we make the assumption that the API returns them in order (that would be nice). This is just a bug then.

Integralist:
That's cool, I'm manually sorting things at the moment to work around it:

```go
package main

import (
	"fmt"
	"os"
	"sort"
	"strconv"

	fastly "github.com/sethvargo/go-fastly"
)

type version struct {
	Number  int
	Version *fastly.Version
}

type wrappedVersions []version

// Satisfy the Sort interface
func (v wrappedVersions) Len() int      { return len(v) }
func (v wrappedVersions) Swap(i, j int) { v[i], v[j] = v[j], v[i] }
func (v wrappedVersions) Less(i, j int) bool {
	return v[i].Number < v[j].Number
}

func main() {
	client, err := fastly.NewClient("123")
	if err != nil {
		fmt.Println(err)
	}

	fastlyServiceID := "456"

	listVersions, err := client.ListVersions(&fastly.ListVersionsInput{
		Service: fastlyServiceID,
	})

	wv := wrappedVersions{}
	for _, v := range listVersions {
		i, err := strconv.Atoi(v.Number)
		if err != nil {
[...]

PR #40, opened 2017-04-14
Fix the sorting of the versions to be based on numbers and not literals

aerostitch:
Currently the sorting of the versions is based on the sorting of literal fields type.
In a literal sorting, 400 is smaller than 9 for example.
This is what's causing the issue #36.
This PR adds a test showing it and the tests fixing it.

### DIFF
--- a/version.go
+++ b/version.go
@@ -7,14 +7,14 @@ import (
 
 // Version represents a distinct configuration version.
 type Version struct {
-	Number    string     `mapstructure:"number"`
-	Comment   string     `mapstructure:"comment"`
-	ServiceID string     `mapstructure:"service_id"`
-	Active    bool       `mapstructure:"active"`
-	Locked    bool       `mapstructure:"locked"`
-	Deployed  bool       `mapstructure:"deployed"`
-	Staging   bool       `mapstructure:"staging"`
-	Testing   bool       `mapstructure:"testing"`
+	Number    int    `mapstructure:"number"`
+	Comment   string `mapstructure:"comment"`
+	ServiceID string `mapstructure:"service_id"`
+	Active    bool   `mapstructure:"active"`
+	Locked    bool   `mapstructure:"locked"`
+	Deployed  bool   `mapstructure:"deployed"`
+	Staging   bool   `mapstructure:"staging"`
+	Testing   bool   `mapstructure:"testing"`
 }
 
 // versionsByNumber is a sortable list of versions. This is used by the version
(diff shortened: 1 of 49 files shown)


version.go
```diff
@@ -25,7 +26,9 @@ type versionsByNumber []*Version
 func (s versionsByNumber) Len() int      { return len(s) }
 func (s versionsByNumber) Swap(i, j int) { s[i], s[j] = s[j], s[i] }
 func (s versionsByNumber) Less(i, j int) bool {
-	return s[i].Number < s[j].Number
```
sethvargo: If we choose to be Go 1.8+, we can shorten this a bit to:

```go
sort.Slice(versions, func(i, j int) bool {
  return versions[i].Number < versions[j].Number
})
```

https://golang.org/pkg/sort/#Slice

What's weird is that Fastly's API clearly says that  "Number" is an integer, yet I tested it with curl and it's most definitely not. I wonder if there's a larger issue here that's fixable using mapstructure's weakdecode. We might be able to decode the string into an int automatically, and then the sort becomes a very simple less function.

aerostitch: There's probably a lot of 1.6 & 1.7 still out there and I'm not sure that dependent tools like Terraform would be very keen on moving to 1.8+ already.

If the api doesn't return an int we probably should reach out to Fastly directly. I haven't noticed that but it might be on older versions of services.

If you're ok with changing this field's type it would indeed make more sense to me. It also sounded weird to me why `Number` was declared as a `string` in the sdk, but then I figured that proposing this change could have a much bigger impact on the other tools relying on this library, that's why I proposed this change.

sethvargo: It looks like we already have weakly-typed input. If you change the type on that struct to an int (or uint) instead of string, it should "just work". Not sure what to do about BC though.

aerostitch:
Hi @sethvargo,

As requested I changed the type of the Number in the Version struct, and then realized that it breaks everything unless we make it coherent and use int for versions everywhere so I had to change a bit more than expected initially.

I think we could do a little refactoring of the check for the service name and the version like a "checkVersion" and "checkService" method on the client, but I wanted to change as little as possible for now to make sure I didn't miss anything 1st.

The PR is updated with the version set to int as in some places you can set it to `-1` to get the value of the latest version (ie in the diff).

sethvargo:
I had to pull down manually to update all the fixtures, but it works! Great work! 😄

aerostitch:
Thanks! :)

Issue #20, opened 2016-07-26
DefaultTTL cannot be set to 0


keshavdv:
In cases where I'd like to disable Fastly caching by default, I would set the default TTL to 0. However, the use of omitempty in the UpdateSettingsInput struct means that the TTL field is omitted entirely if the value is set to 0. A possible solution is to make the field required by removing the omitempty tag or turning DefaultTTL into a pointer. I'm glad to make a PR for either approach but wanted to hear your thoughts.

keshavdv:
It seems that even with some modification, the underlying library results in incorrect behavior (https://github.com/ajg/form/issues/11).

aerostitch:
Proposed a fix in #41

PR #41, opened 2017-04-14
Update vendors + allow default_ttl to be set to 0

aerostitch:
As described in #20 and https://github.com/hashicorp/terraform/pull/13648 and https://github.com/hashicorp/terraform/issues/12910, not being able to set the `default_ttl` to `0` is quite an unexpected behavior.
To be able to do that we have to update the `github.com/ajg/form` dependency (I used `make updatedeps` so all have been updated, tested and had to modify `fastly_test.go` to fix an incompatibility), which is the 1st commit of this PR.

The 2nd removes the omitempty which removes the parameter when set to 0 and change the form encoder to keep zero values. Added a test for it.
Let me know if that's acceptable.

Thanks!


aerostitch:
Weird, the tests don't fail in go 1.8 (including the fix proposed in #39):
```
$ make test testrace
==> Generating...
==> Running tests...
ok  	github.com/sethvargo/go-fastly	0.436s
==> Running tests (race)...
ok  	github.com/sethvargo/go-fastly	2.266s
```
Will dig.

aerostitch:
That was due to go-cleanhttp dropping compatibility with go 1.6 in https://github.com/hashicorp/go-cleanhttp/commit/a45970658e51fea2c41445ff0f7e07106d007617
Kept that out of the dependencies update.
The test work with 1.6 too now.

PR #42, opened 2017-05-12
remove wordpress APIs

thommahoney:
This removes the deprecated wordpress APIs from this library. There are
other references to wordpress for the Service details endpoint which
will remain for the time being.

These endpoints will no longer be available after May 23, 2017.


thommahoney:
Not sure if there's a more compelete testing mechanism but here's what I've done:

```
› make test
==> Testing github.com/sethvargo/go-fastly...
ok  	github.com/sethvargo/go-fastly	0.082s
```

PR #45, opened 2017-07-20
Add support for ACLs and ACL entries

msuterski:
Implements an interface for Fastly's ACLs and ACL entries.

I did not include fixtures but have them all generated locally from my Fastly account.

### DIFF
--- a/acl.go
+++ b/acl.go
@@ -1,0 +1,216 @@
+package fastly
+
+import (
+	"fmt"
+	"sort"
+)
+
+type Acl struct {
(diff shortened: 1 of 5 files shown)


sethvargo:
I just merged this, however, I wanted to get your thoughts on ACL vs Acl. According to go, I _think_ it should be ACL (like URL or HTTP).

msuterski:
You're right. My bad. Sorry.

I'll file a fix for it today. 

Thanks for a quick merge.

PR #48, opened 2017-07-26
Add interface for stats and realtime analytics

msuterski:
This adds an interface to the historical and realtime stats endpoints.

The tests do not verify any data because it would be different for any service or account. Just verify that the http requests are successful.

### DIFF
--- a/fastly/client.go
+++ b/fastly/client.go
@@ -27,6 +27,9 @@ const APIKeyHeader = "Fastly-Key"
 // support an on-premise solution, this is likely to always be the default.
 const DefaultEndpoint = "https://api.fastly.com"
 
+// RealtimeStatsEndpoint is the realtime stats endpoint for Fastly.
+const RealtimeStatsEndpoint = "https://rt.fastly.com"
+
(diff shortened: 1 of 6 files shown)


sethvargo:
This looks really good - one question. Do you think realtimestats should be its own client? It feels a bit weird that these methods would be available on a regular client too. Or is that intended?

msuterski:
The realtime stats are published from a different endpoint (`api.fasty.com` vs `rt.fastly.com`). I'm not sure how we could use the same client to talk to different hostnames? Is there a way?

I actually had it implemented as a separate type to be explicit which methods are available through which client type, but though it was a bit of an overkill. I can put it back.

sethvargo:
Hey @msuterski 

Yea - I know the RTS are different endpoints, which is why I think it makes sense for them to be different API clients.

msuterski:
Ok, I'll push an update to this. Thanks!

msuterski:
Updated to use a separate client type for the RTS API calls.

Let me know if it works for you, please.

Issue #54, opened 2017-10-03
ListWAFs seems unable to unmarshal LastPush into *time.Time 


bridgetlane:
When I use ListWAFs:
```
	var serviceID = "derp"

	latest, err := client.ListWAFs(&fastly.ListWAFsInput{
	 	Service: serviceID,
		Version: 19,
	})
	if err != nil {
		log.Fatal(err)
	 	log.Fatal(latest)
	}
```
I'm getting the following error:
```
2017/10/03 10:43:52 Only numbers can be parsed as dates, unix timestamps
```
It seems to be erroring in the ListWAFs unmarshal, I believe because this UTC date format ("2016-09-28 22:52:33 UTC") that is returned in the field last_push is trying to be unmarshalled into *time.Time.

bridgetlane:
Fastly has pushed an update to fix the UTC formatting issue from what was being returned by their API, so we should be able to use time.Time now.

PR #55, opened 2017-10-05
Add OWASP-functionality

bridgetlane:


bridgetlane:
Added 2/4 of the rules endpoints, the other two are blocked by https://github.com/sethvargo/go-fastly/issues/54 or until I find an ok solution

bridgetlane:
Added rule sets endpoints as well

udryan10:
+1 
@sethvargo, anything holding this up from getting merged?

fastly/waf.go
```diff
@@ -15,11 +14,11 @@ type WAFConfigurationSet struct {
...
-	ID                string     `jsonapi:"primary,waf"`
-	Version           int        `jsonapi:"attr,version"`
-	PrefetchCondition string     `jsonapi:"attr,prefetch_condition"`
-	Response          string     `jsonapi:"attr,response"`
-	LastPush          *time.Time `jsonapi:"attr,last_push"`
+	ID                string `jsonapi:"primary,waf"`
+	Version           int    `jsonapi:"attr,version"`
+	PrefetchCondition string `jsonapi:"attr,prefetch_condition"`
+	Response          string `jsonapi:"attr,response"`
+	LastPush          string `jsonapi:"attr,last_push"`
```
sethvargo: Is there a reason this is a string instead of a time? It appears to be a time value.

bridgetlane: @sethvargo  Yeah, because of issue #54

bridgetlane: I could not get this to properly unmarshal into a time value, so I changed it to a string for the purposes of being able to use this while I was making changes. I'm open to feedback here also

PR #58, opened 2017-11-21
Add support for `placement` when creating a new logging endpoint

RexBelli:
Add support for the `placement` field (described here: https://docs.fastly.com/guides/web-application-firewall/fastly-waf-logging#waf_debug_log) to allow a logging endpoint to be used for waf logs.

### DIFF
--- a/fastly/s3.go
+++ b/fastly/s3.go
@@ -101,6 +101,7 @@ type CreateS3Input struct {
 	ResponseCondition string       `form:"response_condition,omitempty"`
 	TimestampFormat   string       `form:"timestamp_format,omitempty"`
 	Redundancy        S3Redundancy `form:"redundancy,omitempty"`
+        Placement         string        `form:"placement,omitempty"`
 }
 
 // CreateS3 creates a new Fastly S3.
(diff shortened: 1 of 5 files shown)


fastly/s3.go
```diff
@@ -101,6 +101,7 @@ type CreateS3Input struct {
 	ResponseCondition string       `form:"response_condition,omitempty"`
 	TimestampFormat   string       `form:"timestamp_format,omitempty"`
 	Redundancy        S3Redundancy `form:"redundancy,omitempty"`
+        Placement         string       `form:"placement,omitempty"`
```
thommahoney: I'm sure gofmt will be angry about this, but if it's not, could you align with the line above?

RexBelli: I really did try, I'm not sure why the diff is showing that.
https://github.com/RexBelli/go-fastly/blob/fd99546ee9ec7afbac829ca7fe81c087707b2377/fastly/s3.go#L104

thommahoney: weird github display issue!

PR #60, opened 2017-12-20
Methods for manipulating status of WAF rules

rabdill:


rabdill:
Thanks @sethvargo ! Hope the new gig is going well.

PR #66, opened 2018-01-05
Stop double encoding Condition URLs

thomshutt:
This addresses https://github.com/sethvargo/go-fastly/issues/65
  
It looks like the reason URL encoding was being done was to accomodate forward slashes in Condition names. If you look at the test fixtures however, you can see that the slash was being double encoded anyway, so this would never have worked AFAICT.

This fix makes it possible to now use any non-slash special character - I think that how to deal with slashes will need to be a broader fix, as it will affect any element that can be addressed by name in the URL.

### DIFF
--- a/fastly/condition.go
+++ b/fastly/condition.go
@@ -121,7 +121,7 @@ func (c *Client) GetCondition(i *GetConditionInput) (*Condition, error) {
 		return nil, ErrMissingName
 	}
 
-	path := fmt.Sprintf("/service/%s/version/%d/condition/%s", i.Service, i.Version, url.PathEscape(i.Name))
+	path := fmt.Sprintf("/service/%s/version/%d/condition/%s", i.Service, i.Version, i.Name)
 	resp, err := c.Get(path, nil)
 	if err != nil {
 		return nil, err
@@ -163,7 +163,7 @@ func (c *Client) UpdateCondition(i *UpdateConditionInput) (*Condition, error) {
 		return nil, ErrMissingName
 	}
 
-	path := fmt.Sprintf("/service/%s/version/%d/condition/%s", i.Service, i.Version, url.PathEscape(i.Name))
+	path := fmt.Sprintf("/service/%s/version/%d/condition/%s", i.Service, i.Version, i.Name)
 	resp, err := c.PutForm(path, i, nil)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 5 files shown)


sethvargo:
Hi @thomshutt 

I'm fairly certain it's supposed to be encoded, unless something changed in Fastly's API:

https://github.com/sethvargo/go-fastly/issues/14#issuecomment-315172360

When I run the tests, they are actual integration tests that verify the request and response, so Fastly is correctly handling the encoded slash.

thomshutt:
Hmm... why does `url.PathEscape` only get called for the names of Conditions (and not other named config types)?

I've put together a little test case of the problem that I hit. Taking `url.PathEscape` away fixes this case, but breaks the 'forward slash' case:

```
package main

import (
	"github.com/sethvargo/go-fastly/fastly"
)

func main() {
	client := fastly.DefaultClient()

	// Create a service
	service, err := client.CreateService(&fastly.CreateServiceInput{Name: "Name Encoding Test"})
	if err != nil {
		panic(err)
	}

	// Create a condition with a space in the name
	conditionName := "abc 123"
	_, err = client.CreateCondition(&fastly.CreateConditionInput{
		Service:   service.ID,
		Version:   1,
		Name:      conditionName,
		Type:      "request",
		Statement: `req.url.path == "/test.html"`,
	})
	if err != nil {
		panic(err)
	}

	// Try to 'get' that condition - this will return an error
	_, err = client.GetCondition(&fastly.GetConditionInput{
		Service: service.ID,
		Version: 1,
		Name: conditionName,
	})
	if err != nil {
		panic(err)
	}
}

```

This exits with:
```
panic: 404 - Not Found:

    Title:  Record not found
    Detail: Couldn't find Condition '{ version => 1, name => abc%20123, deleted => 0000-00-00 00:00:00, service => 65Kzvlm97GtdYAkh3ZCs03 }'
```

sethvargo:
Hey @thommahoney

Do you know what fields need to be escaped and which do not?

thommahoney:
Given that test failure, I would think that it was getting double-encoded somehow. Condition names in the URL should be URL-encoded (especially if they contain a slash).

I'm interpreting this updated fixture to mean that it is still getting encoded properly even without the explicit call to `url.PathEscape`: https://github.com/sethvargo/go-fastly/pull/66/files#diff-22a168e96196d8f1445796e9e8d7c2d4R23 but my interpretation could be wrong.

thomshutt:
I've done a bit more digging here - the problem is that when creating the request, we're calling `url.String()` here https://github.com/sethvargo/go-fastly/blob/10e52edcd71b88cb15195ee5380f77a149089e94/fastly/request.go#L44

There's some internal logic to that method that won't do what we want unless both Path and RawPath are set correctly. I've created a gist to show the current situation and roughly what a solution will look like:

https://play.golang.org/p/Pri8abyfDh-

sethvargo:
This looks... fun. I wonder if this should be raised upstream - it shouldn't be this hard...

philippschulte:
I'll close this one in favor of #106.

PR #79, opened 2018-08-17
add function to delete multiple dictionary items

tyanko1:


sethvargo:
Hmm. I think this is useful, but this client library tries to mirror the API as much as possible. I'd rather lobby Fastly to add a batch-delete endpoint.

PR #83, opened 2018-08-31
Make Snippet "Type" a type (like Headers, Cache and Request Settings)

hynd:
Instead of a string (just trying to make usage a little more consistent with some of the other API handlers).


hynd:
Closing for now - have another couple of consistency tweaks to propose...

PR #86, opened 2018-09-12
Add a few missing fields to some API's

hynd:
* The CreateDirectorInput and UpdateDirectorInput were missing the
`capacity` field, added the `shield` field.

* Added `name` (NewName) to UpdateDirectorInput, UpdateConditionInput

* Added `write_only` bool to CreateDictionary/UpdateDictionary

* Added `comment` to various, and created_at/deleted_at/updated_at to
  those that support it (eg; those that can be updated dynamically).

* Removed `locked` from Domain (looks like a copypasta from elsewhere?)

### DIFF
--- a/fastly/acl.go
+++ b/fastly/acl.go
@@ -11,6 +11,10 @@ type ACL struct {
 
 	Name string `mapstructure:"name"`
 	ID   string `mapstructure:"id"`
+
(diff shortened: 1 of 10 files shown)


hynd:
Director fixes required for https://github.com/terraform-providers/terraform-provider-fastly/pull/43

PR #89, opened 2018-10-04
Add log placement

d1vious:
forgot to include bigquery and updated a mistake with logentires placement


fastly/bigquery.go
```diff
@@ -19,6 +19,7 @@ type BigQuery struct {
 	UpdatedAt         string `mapstructure:"updated_at"`
 	DeletedAt         string `mapstructure:"deleted_at"`
 	ResponseCondition string `mapstructure:"response_condition"`
+	Placement 		  string `mapstructure:"placement"`
```
sethvargo: Please run `go fmt`

d1vious: 👍 will do

PR #93, opened 2019-01-03
add Main attribute to CreateVCLInput

miyagawa:
This PR adds `Main` attribute to the `CreateVCLInput` struct so that it allows you to create a new VCL object as a main VCL rather than creating it as non-main and then set to main with a separate request.

https://docs.fastly.com/api/config#vcl 
> main | boolean | Set to true when this is the main VCL, otherwise false

### DIFF
--- a/fastly/vcl.go
+++ b/fastly/vcl.go
@@ -136,6 +136,7 @@ type CreateVCLInput struct {
 
 	Name    string `form:"name,omitempty"`
 	Content string `form:"content,omitempty"`
+	Main    bool   `form:"main,omitempty"`
 }
 
 // CreateVCL creates a new Fastly VCL.


fastly/vcl.go
```diff
@@ -134,8 +134,9 @@ type CreateVCLInput struct {
 	Service string
 	Version int
 
-	Name    string `form:"name,omitempty"`
-	Content string `form:"content,omitempty"`
+	Name    string       `form:"name,omitempty"`
+	Content string       `form:"content,omitempty"`
+	Main    *Compatibool `form:"main,omitempty"`
```
thommahoney: I think this can just be `bool`

miyagawa: ok force pushed as 8ecc9d24bfc7dc55d8decadb432a7e467363074c

Issue #102, opened 2019-03-28
Support for version comment


philippschulte:
The client doesn't support to add a comment to the version creation payload. There is a feature request for the `terraform-provider-fastly` [#126](https://github.com/terraform-providers/terraform-provider-fastly/issues/126) to support the `comment` attribute in the request payload. As always the functionality needs to be added to the client first.

PR #103, opened 2019-03-29
Add support for version comment in request payload

philippschulte:
This solves #102.


fastly/fixtures/versions/get.yaml
```diff
@@ -9,24 +9,30 @@ interactions:
...
     method: GET
   response:
-    body: '{"testing":false,"locked":false,"number":696,"active":false,"service_id":"7i6HN3TK9wS159v2gPAZ8A","staging":false,"created_at":"2017-07-20T01:15:13+00:00","deleted_at":null,"comment":"","updated_at":"2017-07-20T01:15:13+00:00","deployed":false}'
+    body: '{"testing":false,"locked":false,"number":16,"active":false,"service_id":"7i6HN3TK9wS159v2gPAZ8A","staging":false,"created_at":"2019-03-29T00:33:36Z","deleted_at":null,"comment":"test
+      comment","updated_at":"2019-03-29T00:33:36Z","deployed":false}'
     headers:
       Accept-Ranges:
       - bytes
+      - bytes
+      - bytes
```
sethvargo: This is probably a bug in the VCR recorder, but this is leaking. No action required on your part, more of a mental note for me.

fastly/version.go
```diff
@@ -84,10 +84,11 @@ func (c *Client) LatestVersion(i *LatestVersionInput) (*Version, error) {
 type CreateVersionInput struct {
 	// Service is the ID of the service (required).
 	Service string
+
+	Comment string `form:"comment,omitempty"`
```
sethvargo: Please comment this (irony recognized) using Go's comment string so it's rendered in the client library docs on godoc.org.

fastly/fixtures/versions/get.yaml
philippschulte: This seems like a bug. I have checked my latest PRs and they all have the same issue with `Accept-Ranges` in the recordings.

fastly/version.go
philippschulte: That makes sense. I'll also add a comment to the `Comment` field of the `UpdateVersionInput` struct.

PR #110, opened 2019-05-23
#109 Removing all the references from the old project location

matpimenta:
This is just a "find and replace" on the repo removing all the references.


philippschulte:
LGTM! However, there will be a conflict with #106. I would recommend that we first merge #106 and then rebase this PR because it's easier this way. @matpimenta what are your thoughts on this?

matpimenta:
agreed, @philippschulte. I can rebase/reapply after #106 is merged

PR #116, opened 2019-05-28
Fix Client.Purge method

chen-anders:
The current `Client.Purge` method returns a 404 regardless what I give it. ~It seems to me that the underlying API for purges may have changed since this was last revisited.~

This PR makes changes to the `Client.Purge` method such that we perform purges in the following manner as documented [here](https://docs.fastly.com/guides/purging/authenticating-api-purge-requests#purging-urls-with-an-api-token).

>  If your website is not configured to use HTTPS, we recommend purging using a POST request with a secure Fastly API URL. This will ensure that your API token in the header is encrypted in transit. The request will look like this: `curl -X POST -H Fastly-Key:FASTLY_API_TOKEN https://api.fastly.com/purge/<your_url_here>`

### DIFF
--- a/fastly/purge.go
+++ b/fastly/purge.go
@@ -1,10 +1,6 @@
 package fastly
 
-import (
-	"fmt"
-	"net/http"
-	"net/url"
-)
+import "fmt"
 
 // Purge is a response from a purge request.
 type Purge struct {
@@ -30,7 +26,7 @@ func (c *Client) Purge(i *PurgeInput) (*Purge, error) {
 		return nil, ErrMissingURL
 	}
 
-	req, err := c.createPurgeRequest(i.URL)
+	req, err := c.RawRequest("POST", "purge/" + i.URL, nil)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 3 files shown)


mkuchen:
Finally found this after half a day of debugging. 👍 for fix, documentation did not explain this well.

chen-anders:
Thanks for the feedback @philippschulte! 

After revisiting the code, I found the actual bug with the existing `Client.Purge` method  where we were inappropriately appending the desired purge URL to `https://api.fastly.com` (e.g. `https://api.fastly.com/https://my-purge-url.com/cat-pic.jpg`) causing the 404s, so I created a separate `PurgeRequest` method to handle the particular case when we want the equivalent of `curl -X PURGE https://my-purge-url.com/cat-pic.jpg`.

In the case where the URL scheme of the purge URL passed in is not `https`, we create a request to `POST` a purge to the fastly API endpoint as recommended. 

A small refactor was done to not duplicate the common code paths where appropriate. A test for the new createPurgeRequest function that contains the logic of which type of request has also been created.

phamann:
Hi @chen-anders @philippschulte, apologies for joining this discussion late.

However I disagree with @philippschulte's points above and would prefer that which switch back to your original implementation in https://github.com/fastly/go-fastly/pull/116/commits/0b1457a693daf3f5aa264297dd278c2c202fb14awith some changes.   

Whilst, yes we do support the `PURGE` method against the URL itself, it is bad practice for us to both recommend its use and use as the reference implementation in the `gofastly` API client. For two reasons:
- As the docs state, if you are using plain HTTP this potentially means sending your token over an unencrypted connection and therefore a secuirty risk we shouldn't be exposing via the client. 
- `PURGE` gets converted into a `FASTLYPURGE` in Fastly before hitting your VCL, this is to allow customers to block/authenticate/change logic of the purge before being issued. Hence why some users may be seeing `404` or `403` via this method depending on the logic in their service configuration.

Given then above, I think it is much better that the client defaults to using the `POST https://api.fastly.com/purge/:url` method instead.

However, your original code should be updated to use the `c.Post()` convenience method instead of `c.RawRequest()` as per your findings above.

chen-anders:
Thanks for the feedback @phamann - the requested changes have been made.

PR #123, opened 2019-06-19
implement the batch dictionary item patch endpoint

mdevilliers:
This PR implements the PATCH /service/service_id/dictionary/dictionary_id/items endpoint.
See https://docs.fastly.com/api/config#dictionary_item_dc826ce1255a7c42bc48eb204eed8f7f

I'm not too sure about adding the fixture based tests - will do if needed.

Cheers,

Mark

### DIFF
--- a/fastly/dictionary_item.go
+++ b/fastly/dictionary_item.go
@@ -1,6 +1,7 @@
 package fastly
 
 import (
+	"errors"
 	"fmt"
 	"net/url"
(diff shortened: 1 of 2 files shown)


philippschulte:
👋thank you so much for contributing and putting so much work into this implementation! At this time we have decided to go with #125. Therefore, I'll close this one out in favor of #125.

PR #125, opened 2019-07-16
Implement batch operations for dictionary items

trentrosenbaum:
The primary focus of this PR is to introduce batch operations, (Create, Update, Upsert and delete) covering dictionary items.  
The dictionary item function ```BatchModifyDictionaryItems``` executes a list of BatchDictionaryItem against the FASTLY API endpoint [PATCH/service/service_id/dictionary/dictionary_id/items
](https://docs.fastly.com/api/config#dictionary_item_dc826ce1255a7c42bc48eb204eed8f7f)

The Batch Operations have been placed in a common location, (fastly.go) because they are also used with the Batch endpoint for ACLS, (coming in a separate PR after this has been merged).

The tests around the dictionary, dictionary items and dictionary items batch have been refactored to aid reading.  The tests covering dictionary items and the batches set up their own services for the test and the tear them down.  The fixtures have been updated to reflect the current request/response accepted by the Fastly API above.


fastly/fastly.go
```diff
@@ -5,6 +5,17 @@ import (
...
 
+type BatchOperation string
+
+const (
+	Create BatchOperation = "create"
+	Update BatchOperation = "update"
+	Upsert BatchOperation = "upsert"
+	Delete BatchOperation = "delete"
+
+	BatchModifyMaximumItems = 1000
```
matpimenta: Can we include a comment explaining this or pointing to the doc that describes this behavior

philippschulte: Unfortunately, this limit isn't documented in our docs. At least I couldn't find it 😞

fastly/fixtures/dictionary_items/get.yaml
```diff
@@ -9,22 +9,25 @@ interactions:
...
+    url: https://api.fastly.com/service/17Cozi1mqUCWuTgyYvipeW/dictionary/2qpiTjYtwgTuLyB2N6vFAq/item/test-dictionary-item
     method: GET
   response:
-    body: '{"updated_at":"2019-05-08T22:01:02Z","service_id":"7i6HN3TK9wS159v2gPAZ8A","deleted_at":null,"item_value":"value","dictionary_id":"2PSSfhzn58nbURuobYGT3I","item_key":"test-dictionary-item","created_at":"2019-05-08T22:01:02Z"}'
+    body: '{"created_at":"2019-07-16T11:30:23Z","item_value":"value","dictionary_id":"2qpiTjYtwgTuLyB2N6vFAq","service_id":"17Cozi1mqUCWuTgyYvipeW","updated_at":"2019-07-16T11:30:23Z","item_key":"test-dictionary-item","deleted_at":null}'
     headers:
       Accept-Ranges:
       - bytes
+      - bytes
+      - bytes
```
philippschulte: I did remove the duplicates all manually! This must be a bug of the recorder! I leave it up to you to clean up all the recodings.

fastly/fixtures/dictionary_items/get.yaml
```diff
@@ -9,22 +9,25 @@ interactions:
...
       Age:
       - "0"
+      - "0"
```
philippschulte: Same here!

fastly/dictionary_item.go
```diff
@@ -189,6 +189,49 @@ func (c *Client) UpdateDictionaryItem(i *UpdateDictionaryItemInput) (*Dictionary
...
+
+	Items []BatchDictionaryItem `json:"items"`
+}
+
+type BatchDictionaryItem struct {
+
+
+	Operation BatchOperation `json:"op"`
+	ItemKey   string         `json:"item_key"`
+	ItemValue string         `json:"item_value"`
```
philippschulte: Why not using github.com/ajg/form as we usually do for input structs?

trentrosenbaum: I used the json encoding/decoding because the documentation for the end point showed that the content type was application/json and not application/x-www-form-urlencoded.

fastly/fastly.go
trentrosenbaum: Agreed I could not find it either.  I could at least put a comment explaining that.

fastly/fixtures/dictionary_items/get.yaml
trentrosenbaum: No worries I will go through a correct these.

fastly/fixtures/dictionary_items/get.yaml
trentrosenbaum: No worries I will go through a correct these.

fastly/fastly.go
```diff
@@ -5,6 +5,17 @@ import (
 	"encoding"
 )
 
+type BatchOperation string
+
+const (
+	Create BatchOperation = "create"
+	Update BatchOperation = "update"
+	Upsert BatchOperation = "upsert"
+	Delete BatchOperation = "delete"
```
phamann: Am slightly concerned about having `fastly` package level constants that are very generic, `Create`, `Update` etc which could cause conflict in the future. Could we consider renaming to something more specific?

trentrosenbaum: That is a good point 'BatchOpsCreate' or 'BatchOperationCreate' could be used.  I think the second one is more clear.

fastly/fastly.go
trentrosenbaum: Changed to name to BatchModifyMaximumOperations to reflect its use between dict items and acl entries.   Operations are used over both.  Also added comments to explain it is not currently documented in the API.

fastly/fixtures/dictionary_items/get.yaml
trentrosenbaum: Removed the duplicate entries in the newly created fixtures.

fastly/fixtures/dictionary_items/get.yaml
trentrosenbaum: Removed the duplicate entries in the newly created fixtures.

fastly/fastly.go
trentrosenbaum: Appended the name of BatchOperation to the next of each constant.

PR #134, opened 2019-10-17
Add fastly waf related fields

KeisukeYamashita:
## What

* Add ...
    * `vcl` field in `RuleStatus` struct 
    * `disabled` in `WAF` struct
* Change ...
    * Reorder `RuleStatus` structs field to follow document order

## Why

* Since this is missing compared to [WEB API Doc](https://docs.fastly.com/api/waf)


tehviking:
Hi there! Thank you so much for taking time to note and add missing fields to the WAF API. Currently we're working on some significant updates to this part of the library, specifically around the WAF, so I'd like to close this PR for now. If you have any additional questions, please let me know.

PR #142, opened 2020-01-10
Add dynamic servers support

asobrien:
This adds support for the API's [dynamic servers](https://docs.fastly.com/api/dynamicservers) to the client via the `Pool` and `Server` resources.


asobrien:
The fixtures for the `pool` and `server` resources are currently absent--I wasn't sure what the process around generating looks like.

fastly/pool.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+	sort.Stable(poolsByName(ps))
+	return ps, nil
+}
+
+// CreatePoolInput is used as input to the CreatePool function.
+type CreatePoolInput struct {
+	// Service is the ID of the service. Version is the specific configuration
+	// version. Both fields are required.
+	Service string
```
phamann: We have a lot of inconsistency with `Service` vs `ServiceID` in the client (which we hope to address this year), and it seems you also have inconsistency in this file (`Pool` struct on line 25 uses `ServiceID`). Therefore, can you please change this to `ServiceID`.

(Similar comments elided for the rest of the review)

fastly/pool.go
```diff
@@ -0,0 +1,293 @@
...
+	TLSClientCert    string       `form:"tls_client_cert,omitempty"`
+	TLSSNIHostname   string       `form:"tls_sni_hostname,omitempty"`
+	TLSCheckCert     bool         `form:"tls_check_cert,omitempty"`
+	TLSCertHostname  string       `form:"tls_cert_hostname,omitempty"`
+	MinTLSVersion    string       `form:"min_tls_version,omitempty"`
+	MaxTLSVersion    string       `form:"max_tls_version,omitempty"`
+	Healthcheck      string       `form:"healthcheck,omitempty"`
+	Type             PoolType     `form:"type,omitempty"`
+	OverrideHost     string       `form:"override_host,omitempty"`
+}
```
phamann: Not an issue with your implementation, as we have this problem throughout the project. However there is a problem with using scalar types directly in `UpdateFooInput` input structs, and we intend to move all other structs to this pattern over this year.

The problem is the combination of Go's type zero values, JSON marshalling with the omitempty struct field tag, and a users intent to actually set a zero value. This blog post does a much better job at explaining the problem and well worth a read: https://willnorris.com/2014/05/go-rest-apis-and-pointers/

So in your example, if `UpdatePoolInput.TLSCheckCert` is currently `true` and the user attempts to update the value to `false`, omitempty considers this value empty and won't send the updated value.

As the post suggests, the more robust solution would be to change the struct field values to all be pointers instead of the underlying type value. However (whilst we think this is the correct approach) it adds some pain for the library users to convert their inputs to pointers and do nil pointer checks when consuming etc.

Therefore, what are your thoughts on changing all these values to pointers?

fastly/pool.go
```diff
@@ -0,0 +1,293 @@
...
+	Name string `form:"name"`
+
+	Comment          string       `form:"comment,omitempty"`
+	Shield           string       `form:"shield,omitempty"`
+	RequestCondition string       `form:"request_condition,omitempty"`
+	MaxConnDefault   uint         `form:"max_conn_default,omitempty"`
+	ConnectTimeout   uint         `form:"connect_timeout,omitempty"`
+	FirstByteTimeout uint         `form:"first_byte_timeout,omitempty"`
+	Quorum           uint         `form:"quorum,omitempty"`
+	UseTLS           *Compatibool `form:"use_tls,omitempty"`
```
phamann: Are you sure this needs to be a `Compatibool`?

fastly/server.go
```diff
@@ -0,0 +1,244 @@
...
+// ServerOptionsInput is used to define the optional fields for a server. It is
+// used as part of create and update server operations.
+type ServerOptionsInput struct {
+	Comment      string `form:"comment,omitempty"`
+	Weight       uint   `form:"weight,omitempty"`
+	MaxConn      uint   `form:"max_conn,omitempty"`
+	Port         uint   `form:"port,omitempty"`
+	Disabled     bool   `form:"disabled,omitempty"`
+	OverrideHost string `form:"override_host,omitempty"`
+}
```
phamann: This doesn't seem to be used anywhere, please delete if so?

asobrien:
Thanks for the comments!

> With regards to testing, you'll need to run `make test` with a valid API token, which will execute the requests against the API and record the results (with the API token redacted) for subsequent runs.

I don't have API credentials to access to the [baked-in test service](https://github.com/fastly/go-fastly/blob/master/fastly/fastly_test.go#L18-L19). Do maintainers generally run the tests against this service to generate the fixtures? Or, alternatively should a contributor run the test against a Fastly service they have access to and then update the fixtures accordingly? I'm just trying to get a handle on the process around this.

The only issue with running the tests against an "unofficial" service, is that the version number recorded in the responses become somewhat arbitrary. Just noting that if this is the recommended route, it may be worth allowing the test service ID to be defined via an env var and adding a recorder Filter that modifies the request/response to keep a consistent service ID.

> Also take a look at the helpers @devkitx added in his version of this in #143, specifically here: https://github.com/fastly/go-fastly/pull/143/files#diff-997cff4762ab0b7c60bcaf50fcc36d6e.

I've included nearly identical [helpers](https://github.com/fastly/go-fastly/pull/142/files#diff-997cff4762ab0b7c60bcaf50fcc36d6eR185-R218). I don't have a baked in [`testPoolID `](https://github.com/fastly/go-fastly/pull/143/files#diff-997cff4762ab0b7c60bcaf50fcc36d6eR22) as Pools are created as needed via the `createTestPool` helper. Not referencing a hard-coded testServiceID also sidesteps the issue of the tests not being portable between services as the Pool is tied to a specific service. Maybe I'm missing something?

fastly/server.go
asobrien: Good catch! This should have been removed. I was attempting use a single struct to define common fields between the Server create and update operations but I don't think the serializer supports using embedded structs.

Removed via e088bc5

fastly/pool.go
asobrien: I'm not sure that this needs to be a `Compatibool`, but the API documentation for the [Pool resource](https://docs.fastly.com/api/dynamicservers#api-section-pool) suggests that this should be a [Compatibool](https://github.com/fastly/go-fastly/blob/e18f1a520c404b08eb1d2c219b9010de494aab47/fastly/fastly.go#L48-L50):

>field: use_tls 	
>type: string 	
>description: Whether or not to use TLS to reach any server in the pool. Optional. Defaults to "0".

so that this is marshalled to a "0/1" rather than "true/false". Let me know if that's not the case.

fastly/pool.go
asobrien: I tried to keep consistent usage as the rest of the library, the pattern that stood out to me was the use of `ServiceID` in the API response structs for the resources and `Service` in the related `*Input` structs. For example:

* [`ACLEntry`](https://github.com/fastly/go-fastly/blob/e18f1a520c404b08eb1d2c219b9010de494aab47/fastly/acl_entry.go#L10)
    - [`ListACLEntriesInput`](https://github.com/fastly/go-fastly/blob/e18f1a520c404b08eb1d2c219b9010de494aab47/fastly/acl_entry.go#L35)
    - [`GetACLEntryInput`](https://github.com/fastly/go-fastly/blob/e18f1a520c404b08eb1d2c219b9010de494aab47/fastly/acl_entry.go#L68)

* [`Director`](https://github.com/fastly/go-fastly/blob/e18f1a520c404b08eb1d2c219b9010de494aab47/fastly/director.go#L29)
   - [`CreateDirectorInput`](https://github.com/fastly/go-fastly/blob/e18f1a520c404b08eb1d2c219b9010de494aab47/fastly/director.go#L91)
   - [`DeleteDirectorInput`](https://github.com/fastly/go-fastly/blob/e18f1a520c404b08eb1d2c219b9010de494aab47/fastly/director.go#L214)

A quick set of grep heuristics seems to verify this pattern:

    $  grep -ER '\s+ServiceID\s+string.*' . | wc -l
        36
    $ grep -ER '\s+Service\s+string.*' . | wc -l
        180

Given there are 5 operations per resource (list, get, create, update, delete), this works out surprisingly well.

If you're planning on moving the CRUD operation inputs to use a `ServiceID` field and want to start with these resources I'll happily make the changes. Let me know what you think.

fastly/pool.go
asobrien: I'm definitely in favor of using the pointers in the struct fields. It's true that it makes it a bit more cumbersome for users to set the fields, though a set of helpers (cf. [goprotobuf](https://github.com/golang/protobuf/blob/ed6926b37a637426117ccab59282c3839528a700/proto/lib.go#L457-L461) and [aws](https://github.com/aws/aws-sdk-go/blob/v1.28.5/aws/convert_types.go#L182-L185)) helps. 

I suspect the response structs the lib would want to also use pointers for fields (if nothing else, to keep consistency between inputs and outputs), though it may not be strictly necessary since it seems like the Fastly API always returns all fields in the response struct.

Without using pointers for field structs, there are definitely cases where resources won't be able to be modified, for the Pool resource alone one won't be able to:

* Disable quorom once enabled
* Remove TLSCaCert
* Remove TLSCiphers
* Remove TLSClientKey
* Remove TLSClientCert
* Remove TLSSNIHostname
* Disable Healthcheck
* Remove Comment
* Remove OverrideHost
* Disable TLSCheckCert

I think that list alone makes a compelling case for using pointer fields. Updating the library throughout is undoubtedly a breaking change, but if you want to pave the way with these new resources I'm all for it.

phamann: > I think that list alone makes a compelling case for using pointer fields. Updating the library throughout is undoubtedly a breaking change, but if you want to pave the way with these new resources I'm all for it.

Apologies for the confusion, yes I'd like to start this process for these new resources. You don't need to worry about updating all of the existing ones :)

fastly/pool.go
phamann: Thanks for checking, I'm happy leaving it as a `Compatibool` 👍

fastly/pool.go
asobrien: 👍 Let's do it! I'll make the changes.

asobrien:
@phamann I've updated this PR to use pointers for optional fields in structs (see 
cf93538). I limited the use of pointers to optional fields as this seemed like a good compromise between the existing library and the required functionality.

It would be possible for all the fields in the input structs to be pointers, but this requires adding new nil pointer checks (e.g., [here](https://github.com/asobrien/go-fastly/blob/feature/dynamic_servers/fastly/pool.go#L130)) that differs from the rest of the library. These value fields are also required in each input struct, so there's no real need from them to be pointers. I did modify the Pool update test case to ensure we can actually set a "zero" value (see [here](https://github.com/asobrien/go-fastly/blob/feature/dynamic_servers/fastly/pool_test.go#L114) and [here](https://github.com/asobrien/go-fastly/blob/feature/dynamic_servers/fastly/pool_test.go#L123-L125))

A didn't modify the response structs to use pointer fields either. My thoughts around this are: keeping these as value fields keeps the response consistent with the rest of the library (no need to start dereferencing the reponses). Moreover, as far as I can tell, the Fastly API always populates all the fields in the response which means that we don't need to rely on pointer fields to omit missing fields.

I added some helper functions (cf. 2767792) to make it easier for a user to set optional fields. I also generated the fixtures (and replaced the test service id to match that baked into the test client).

Let me know what you think about how optional fields are being handled in the Pool and Server resources and if that's what you had in mind.

fastly/pool.go
phamann: Ok I'm happy to leave as is and we'll do the cleanup with the wider refactor later in the year 👍

fastly/pool.go
```diff
@@ -0,0 +1,293 @@
...
+	ConnectTimeout   *uint        `form:"connect_timeout,omitempty"`
+	FirstByteTimeout *uint        `form:"first_byte_timeout,omitempty"`
+	Quorum           *uint        `form:"quorum,omitempty"`
+	UseTLS           *Compatibool `form:"use_tls,omitempty"`
+	TLSCACert        *string      `form:"tls_ca_cert,omitempty"`
+	TLSCiphers       *string      `form:"tls_ciphers,omitempty"`
+	TLSClientKey     *string      `form:"tls_client_key,omitempty"`
+	TLSClientCert    *string      `form:"tls_client_cert,omitempty"`
+	TLSSNIHostname   *string      `form:"tls_sni_hostname,omitempty"`
+	TLSCheckCert     *bool        `form:"tls_check_cert,omitempty"`
```
philippschulte: I think `TLSCheckCert` should be a `Compatibool` as well. The API does return a string `"0"/"1"` and not `false/true`.

fastly/pool.go
```diff
@@ -0,0 +1,293 @@
...
+	TLSClientKey     string   `mapstructure:"tls_client_key"`
+	TLSClientCert    string   `mapstructure:"tls_client_cert"`
+	TLSSNIHostname   string   `mapstructure:"tls_sni_hostname"`
+	TLSCheckCert     bool     `mapstructure:"tls_check_cert"`
+	TLSCertHostname  string   `mapstructure:"tls_cert_hostname"`
+	MinTLSVersion    string   `mapstructure:"min_tls_version"`
+	MaxTLSVersion    string   `mapstructure:"max_tls_version"`
+	Healthcheck      string   `mapstructure:"healthcheck"`
+	Type             PoolType `mapstructure:"type"`
+	OverrideHost     string   `mapstructure:"override_host"`
```
philippschulte: Please add `created_at`, `updated_at`, and `deleted_at` fields to the response struct. These fields are not listed in the docs but belong to the response body.

fastly/pool.go
philippschulte: I think it is important that `TLSCheckCert` and `UseTLS` are of type `Compatibool` in the `Pool` struct declaration.

asobrien: I can confirm that `TLSCheckCert` returns "0/1" rather than "true/false", I'll update this field to use a `Compatibool`.

The [API docs](https://docs.fastly.com/api/dynamicservers#api-section-pool) note that the default is "0" for `UseTLS`, however the response style and default for `TLSCheckCert` isn't noted in the description.

asobrien: Fixed in 59ec1eb

fastly/pool.go
asobrien: Fixed in 25093b6

philippschulte:
@asobrien thanks for addressing the PR comments but I can't find 59ec1eb, 816dd2b, and 25093b6. Where did they go 🤔

asobrien:
Sorrty @philippschulte forgot to push before I got distracted. Should be there now.

philippschulte:
@asobrien looks good but also please update `UseTLS` and `TLSCheckCert` to `* Compatibool` in the response struct too. After that, I can merge and cut a new release.

asobrien:
@philippschulte I'm not sure there's a way to get rid of pointers in the input. Looking into the [KeepZeros](https://github.com/ajg/form#zero-values) option in `github.com/ajg/form`, it looks that option ensure that zero-values will be submitted as a native-type (desired behavior) rather than an a empty string. Without using pointers for optional inputs, it's impossible to set/revert/update and field to be represented by the zero-value as this will get dropped during marshalling as a result of the `omitempty` tag. I tested this behavior on e088bc5 and it isn't possible to set `Quorum = 0`, for example.

Getting rid of the `omitempty` tag allows the zero-values to be set, but every update will overwrite any fields that aren't explicitly defined--I'm not sure this is desirable either. Pointers the only way I know of that allows for optional fields to be optional and for zero-values to be used.

@phamann I looked into the API response (for the Pool resource) in a little more depth, and it seems like the Fastly API doesn't explicitly set unset fields to the default value but rather sets them to `null`, for example:

```
$ curl -sS -H "Fastly-Key: ${FASTLY_API_KEY}" -X GET "https://api.fastly.com/service/deadbeef/version/48/pool" | jq
[
  {
    "name": "mypool",
    "first_byte_timeout": "15000",
    "tls_ciphers": null,
    "type": "random",
    "healthcheck": null,
    "max_conn_default": "100",
    "service_id": "deadbeef",
    "tls_check_cert": "0",
    "created_at": "2020-01-29T13:57:10Z",
    "quorum": "0",
    "request_condition": null,
    "tls_sni_hostname": null,
    "deleted_at": null,
    "min_tls_version": null,
    "shield": null,
    "tls_cert_hostname": null,
    "between_bytes_timeout": "10000",
    "version": "48",
    "max_tls_version": null,
    "use_tls": "0",
    "tls_ca_cert": null,
    "updated_at": "2020-01-29T18:55:20Z",
    "id": "abcdef0123456789",
    "tls_client_cert": null,
    "connect_timeout": "1000",
    "tls_client_key": null,
    "comment": "",
    "override_host": null
  }
]
```

without using pointers in the response struct, it's not possible to tell whether those fields are _unset_ (`"shield": null`) or are just set the default value (`"shield": ""`).

asobrien:
> please update `UseTLS` and `TLSCheckCert` to `* Compatibool` in the response struct too

Fixed in 621dd64

philippschulte:
@asobrien that is exactly what I found out about the `KeepZeros` option too 😆. I appreciate that you also looked into it because I was thinking I am not using it correctly. @phamann told me this morning that he is fine whit the pointer solution and that it will be the new standard of the library!

For consistency please use a pointer for  `Compatibool` in the response struct (621dd64). I am happy to merge the PR and cut a new release after this has been addressed. Thanks

asobrien:
@philippschulte Will do! I went through the library, and it looks like [none of the response structures](https://github.com/fastly/go-fastly/search?utf8=%E2%9C%93&q=UseTLS&type=) currently use a `Compatibool` type (they all use `bool`), for example in [logentries](https://github.com/fastly/go-fastly/blob/78021c650d98cc7ba73622600c3f67e68722fcaf/fastly/logentries.go#L17), [syslog](https://github.com/fastly/go-fastly/blob/50792f54ba0a4b83953dc993f24ffe4095a5844e/fastly/syslog.go#L19) and [backend](https://s3.console.aws.amazon.com/s3/buckets/giphy-etleap/?region=us-east-1&tab=permissions). I can confirm the response correctly unmarshals into a `bool`.

In this race, I've got no horse. I just wanted to point it out since it makes a break with the existing response structs.

asobrien:
No problem. Just to clarify: we should use a `bool` in the `Pool` response struct?

If yes, let me revert and clean up the last commit and then we should be good to go.

philippschulte:
@asobrien yes please revert it! My fault, sorry 😭

asobrien:
@philippschulte fixed in 0f6b649

philippschulte:
@asobrien [v1.5.0](https://github.com/fastly/go-fastly/releases/tag/v1.5.0) has been released! Thanks again.

PR #143, opened 2020-01-14
Add dynamic server support

devkitx:
Hello, 
I have added pool and server functionality in order to support dynamic servers as outlined https://docs.fastly.com/api/dynamicservers. The tests for pool and server all pass with make test, with further testing being performed against live Fastly services that have dynamic servers enabled.
Please let me know if there are any changes as I am more than happy to iterate with this.
Thank you
Ed


phamann:
Hi @devkitx thank you for taking your time to add this contribution, we really appreciate it. 

Unfortunately it seems @asobrien submitted a nearly identical PR (#142) a couple of days earlier than yours (which is pretty spooky), so if it's ok with you I'm going to close this one to avoid repetition and focus our efforts on that review?

I'd really appreciate it if you could give that PR a review, so we can ensure you're also happy and it meets your companies needs/requirements.

Once again, thank you for your contribution 😄

PR #146, opened 2020-03-20
Serialize modifications.

cgeisser:
Serialize fastly client operations that make modifications (everything except `HEAD` and `GET` requests).

### DIFF
--- a/fastly/client.go
+++ b/fastly/client.go
@@ -51,8 +51,8 @@ type Client struct {
 	// client will be used.
 	HTTPClient *http.Client
 
-	// updateLock forces serialization of calls that modify a service to
-	// avoid rate limits.
+	// updateLock forces serialization of calls that modify a service.
+	// Concurrent modifications have undefined semantics.
 	updateLock sync.Mutex
 
 	// apiKey is the Fastly API key to authenticate requests.
--- a/fastly/client.go
+++ b/fastly/client.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"runtime"
 	"strings"
+	"sync"
 
 	"github.com/ajg/form"
 	"github.com/google/jsonapi"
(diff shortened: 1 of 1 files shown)


fastly/client.go
```diff
@@ -198,7 +203,13 @@ func (c *Client) Request(verb, p string, ro *RequestOptions) (*http.Response, er
 		return nil, err
 	}
 
+	if verb != "GET" && verb != "HEAD" {
+		c.updateLock.Lock()
+	}
 	resp, err := checkResp(c.HTTPClient.Do(req))
+	if verb != "GET" && verb != "HEAD" {
+		c.updateLock.Unlock()
```
thommahoney: Does it make sense to `defer c.updateLock.Unlock()` in the conditional above?

fastly/client.go
```diff
@@ -198,7 +203,13 @@ func (c *Client) Request(verb, p string, ro *RequestOptions) (*http.Response, er
...
+	}
```
phamann: We can avoid the need for this additional block and risk the conditions getting out of sync by using a `defer` statement directly after your `.Lock()` above. Such as:

```
if verb != "GET" && verb != "HEAD" {
		c.updateLock.Lock()
		defer c.updateLock.Unlock()
}
```

phamann:
@philippschulte If you're happy with the above can we merge and cut a new release?

fastly/client.go
```diff
@@ -198,7 +203,13 @@ func (c *Client) Request(verb, p string, ro *RequestOptions) (*http.Response, er
 		return nil, err
 	}
 
+	if verb != "GET" && verb != "HEAD" {
```
thommahoney: I wonder if it makes sense to use an inclusion list instead of exclusions. This might be missing `PURGE`, for example.

```suggestion
	if verb == "POST" || verb == "PUT" || verb == "PATCH" || verb == "DELETE" {
```

cgeisser: I prefer exclusion lists for things like this so it fails safe (more locking) if I've missed something.

PR #147, opened 2020-03-27
Let purges run in parallel.

cgeisser:
Allow purges to bypass the lock to on modifications.


chen-anders:
Verified that things are fast again on my end. Thanks!

PR #148, opened 2020-03-29
Update travis config to include v1.14.1

chenrui333:


.travis.yml
```diff
@@ -1,13 +1,12 @@
...
 
 go:
-  - "1.10"
-  - "1.11"
-  - "1.12"
-  - "1.13"
+  - "1.10.x"
+  - "1.11.x"
+  - "1.12.x"
+  - "1.13.x"
```
chenrui333: this would include the latest release for 1.{10,11,12,13}

go.mod
```diff
@@ -10,3 +10,5 @@ require (
 	gopkg.in/check.v1 v1.0.0-20180628173108-788fd7840127 // indirect
 	gopkg.in/yaml.v2 v2.0.0-20170812160011-eb3733d160e7 // indirect
 )
+
+go 1.13
```
sbfaulkner, line 14: @chenrui333 @phamann is this intentional that 1.13 is now explicitly required?

chenrui333: That is good call out, as I can see the PR builds were all successful, it should be less a concern here though.

sbfaulkner: I'm getting failures when trying to use it in `terraform-provider-fastly`
https://github.com/terraform-providers/terraform-provider-fastly/pull/214

I'm definitely not an expert on the cross-version compatibility of go modules though

chenrui333: I saw the provider is also using go 1.13?
https://github.com/terraform-providers/terraform-provider-fastly/blob/master/go.mod#L10

chenrui333: what exact error did you refer to?

sbfaulkner: in case anyone is following along at home... https://github.com/terraform-providers/terraform-provider-fastly/pull/214#issuecomment-607486563

PR #153, opened 2020-04-14
WAF-1740 use content-type header for bulk requests

squeemishly:
This change addresses the change made to use the `content-type` header to process bulk requests on the WAF API:

https://github.com/fastly/moonbeam/pull/825
https://jira.corp.fastly.com/browse/WAF-1740

It appears that this API call was added by Oscar when he was creating the WAF provider for Fastly:

https://github.com/fastly/go-fastly/blame/service-waf-implementation/fastly/client.go#L293-L313

It shouldn't impact any other service using go-fastly

### DIFF
--- a/fastly/client.go
+++ b/fastly/client.go
@@ -298,7 +298,7 @@ func (c *Client) RequestJSONAPIBulk(verb, p string, i interface{}, ro *RequestOp
 	if ro.Headers == nil {
 		ro.Headers = make(map[string]string)
 	}
-	ro.Headers["Content-Type"] = jsonapi.MediaType
+	ro.Headers["Content-Type"] = jsonapi.MediaType + "; ext=bulk"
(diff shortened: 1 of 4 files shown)


fastly/client.go
```diff
@@ -298,7 +298,7 @@ func (c *Client) RequestJSONAPIBulk(verb, p string, i interface{}, ro *RequestOp
 	if ro.Headers == nil {
 		ro.Headers = make(map[string]string)
 	}
-	ro.Headers["Content-Type"] = jsonapi.MediaType
+	ro.Headers["Content-Type"] = jsonapi.MediaType + "; ext=bulk"
```
squeemishly, line 301: @bernabas I just added the bulk param to content-type. I didn't remove it from Accept. I'm waiting to hear back in the Moonbeam PR how we want to handle that before I make that change.

PR #154, opened 2020-04-15
Add support for Platform TLS

shreddedbacon:
Looking to add support for Platform TLS

For the test process described [here](https://github.com/fastly/go-fastly/pull/122#discussion_r331670230), I wasn't sure how to actually do this with Platform TLS. 
I don't appear have the ability to create new TLS configuration IDs to have one to use as a dummy, and I am unsure if doing anything here will have any cost implications.


fastly/platform_tls.go
```diff
@@ -0,0 +1,365 @@
...
+}
+
+// BulkCertificateResponseAttribute .
+type BulkCertificateResponseAttribute struct {
+	Type string `json:"type"`
+	ID   string `json:"id"`
+}
+
+// BulkCertificateResponsRelationships .
+type BulkCertificateResponsRelationships struct {
```
gschorkopf, line 194:
```suggestion
type BulkCertificateResponseRelationships struct {
```

fastly/platform_tls.go
```diff
@@ -0,0 +1,390 @@
...
+
+	var ppkr *PlatformPrivateKeysResponse
+	if err := decodeJSON(&ppkr, r.Body); err != nil {
+		return nil, err
+	}
+
+	return ppkr, nil
+}
+
+// GetPlatformPrivateKey returns a specific private key
```
phamann, line 76:
```suggestion
// GetPlatformPrivateKey returns a specific private key.
```

fastly/platform_tls.go
```diff
@@ -0,0 +1,390 @@
...
+func (c *Client) GetPlatformPrivateKey(i *GetPlatformPrivateKeyInput) (*PlatformPrivateKeyResponse, error) {
+
+	p := "/tls/private_keys"
+
+	if i.ID != "" {
+		p = fmt.Sprintf("%s/%s", p, i.ID)
+	}
```
phamann, line 83: We should error early if no `ID` is provided and not sent the request. 

```suggestion
	if i.ID == "" {
		return nil, ErrMissingID
	}
	
	p := fmt.Sprintf("/tls/private_keys/%s", i.ID)
```

fastly/platform_tls.go
```diff
@@ -0,0 +1,390 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/google/jsonapi"
+)
+
+// GetPlatformPrivateKeysInput is an input to the GetPlatformPrivateKeys function.
+// Allowed values for the fields are described at https://docs.fastly.com/api/platform-tls
```
phamann, line 11:
```suggestion
// Allowed values for the fields are described at https://docs.fastly.com/api/platform-tls.
```

fastly/platform_tls.go
```diff
@@ -0,0 +1,390 @@
...
+
+// PlatformPrivateKeysAttributes .
+type PlatformPrivateKeysAttributes struct {
+	KeyLength     string     `mapstructure:"key_length"`
+	KeyType       string     `mapstructure:"key_type"`
+	Name          string     `mapstructure:"name"`
+	CreatedAt     *time.Time `mapstructure:"created_at"`
+	Replace       bool       `mapstructure:"replace"`
+	PublicKeySHA1 string     `mapstructure:"public_key_sha1"`
+}
```
phamann, line 54: As this is a JSONAPI style API format, we can use the [`jsonapi`](https://github.com/google/jsonapi) package to help marshal/unmarshal JSONAPI responses. Therefore, you shouldn't need to model the response data as three separate structs like this as the information we really care about which models a "private key" is the attributes field of the response. Therefore, `PlatformPrivateKeyAttributes` should become `PlatformPrivateKey` and the rest should be removed. You will need to use `jsonapi:""` field tags instead of the `mapstructure` ones.

fastly/platform_tls.go
```diff
@@ -0,0 +1,390 @@
...
+		Params: map[string]string{},
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	var ppkr *PlatformPrivateKeysResponse
+	if err := decodeJSON(&ppkr, r.Body); err != nil {
+		return nil, err
+	}
```
phamann, line 71: As per comments above, this is a JSONAPI response and therefore we should unmarshal using the [`jsonapi`](https://github.com/google/jsonapi) package so you don't need to model the response envelope.  The library provides two helpers, the singular `UnmarshalPayload` and `UnmarshalManyPayload` for multiple. We have to use reflection for the many payload to let jsonapi know the type its decoding into. More info here: https://github.com/google/jsonapi#unmarshalmanypayload

```suggestion
	data, err := jsonapi.UnmarshalManyPayload(resp.Body, reflect.TypeOf(new(PlatformPrivateKey)))
	if err != nil {
		return nil, err
	}
```

fastly/platform_tls.go
```diff
@@ -0,0 +1,390 @@
...
+	KeyLength     string     `mapstructure:"key_length"`
+	KeyType       string     `mapstructure:"key_type"`
+	Name          string     `mapstructure:"name"`
+	CreatedAt     *time.Time `mapstructure:"created_at"`
+	Replace       bool       `mapstructure:"replace"`
+	PublicKeySHA1 string     `mapstructure:"public_key_sha1"`
+}
+
+// GetPlatformPrivateKeys returns stats data based on GetPlatformPrivateKeysInput
+func (c *Client) GetPlatformPrivateKeys() (*PlatformPrivateKeysResponse, error) {
```
phamann, line 57: To follow convention in the rest of the go-fastly package, I'd prefer it if this is called `ListPlatformPrivateKeys()` or similar. The `/tls/private_keys` endpoint is also paginated, so this method should accept a ListPrivateKeysInput struct which models the input parameters `FilterInUse`, `PageNumber` and `PageSize`.

fastly/platform_tls.go
```diff
@@ -0,0 +1,390 @@
...
+
+	p := "/tls/private_keys"
+
+	r, err := c.Get(p, &RequestOptions{
+		Params: map[string]string{},
+	})
```
phamann, line 63: As above it would be good to pass request params for pagination here from an input struct.

fastly/platform_tls.go
```diff
@@ -0,0 +1,390 @@
...
+
+	p := "/tls/private_keys"
+
+	if i.ID != "" {
+		p = fmt.Sprintf("%s/%s", p, i.ID)
+	}
+
+	r, err := c.Get(p, &RequestOptions{
+		Params: map[string]string{},
+	})
```
phamann, line 87: You can pass `nil` instead of an empty `RequestOptions`.
```suggestion
	r, err := c.Get(p, nil)
```

fastly/platform_tls.go
```diff
@@ -0,0 +1,390 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	var ppkr *PlatformPrivateKeyResponse
+	if err := decodeJSON(&ppkr, r.Body); err != nil {
+		return nil, err
+	}
```
phamann, line 95: As per comments above, we can use the jsonapi unmarshal methods here:
```suggestion
	var ppk PlatformPrivateKey
	if err := jsonapi.UnmarshalPayload(resp.Body, &ppk); err != nil {
		return nil, err
	}
```

fastly/platform_tls.go
```diff
@@ -0,0 +1,390 @@
...
+type CreatePlatformPrivateKeyData struct {
+	Type       string                             `json:"type"`
+	Attributes CreatePlatformPrivateKeyAttributes `json:"attributes"`
+}
+
+// CreatePlatformPrivateKeyAttributes .
+type CreatePlatformPrivateKeyAttributes struct {
+	Key  string `json:"key,omitempty"`
+	Name string `json:"name,omitempty"`
+}
```
phamann, line 116: As above, let's just model the attributes and delete the JSONAPI structs. So `CreatePlatformPrivateKeyAttributes` should become `CreatePlatformPrivateKeyInput `

fastly/platform_tls.go
```diff
@@ -0,0 +1,390 @@
...
+
+// CreatePlatformPrivateKey returns a specific private key
```
phamann, line 118:
```suggestion
// CreatePlatformPrivateKey returns a specific private key.
```

fastly/platform_tls.go
```diff
@@ -0,0 +1,390 @@
...
+
+	r, err := c.PostJSON(p, i, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	var ppkr *PlatformPrivateKeyResponse
+	if err := decodeJSON(&ppkr, r.Body); err != nil {
+		return nil, err
+	}
```
phamann, line 139:
```suggestion
        var ppk PlatformPrivateKey
	if err := jsonapi.UnmarshalPayload(resp.Body, &ppk); err != nil {
		return nil, err
	}
```

shreddedbacon:
Brilliant, thanks. I will work through this and let you know when ready.

shreddedbacon:
Hi @phamann. I've adjusted my code to use the `jsonapi` style format as requested.
Hopefully I've got everything correct.
Thanks!

scheying:
Hey guys! Great to see progress on the TLS API. I have been working on the "Non-Platform" TLS part and from what I see the Private Keys API is the same for "Regular TLS" and "Platform TLS". Would it make sense to give the key-related functions neutral names like `GetPrivateKey` instead of `GetPlatformPrivateKey` and move all the key related functionality to a separate file?

If you like I could throw in my implementation of exactly that and a bit of JSONAPI paging and filtering on top. But I also don't want to get in the way of what you already did @shreddedbacon .

shreddedbacon:
> Hey guys! Great to see progress on the TLS API. I have been working on the "Non-Platform" TLS part and from what I see the Private Keys API is the same for "Regular TLS" and "Platform TLS". Would it make sense to give the key-related functions neutral names like `GetPrivateKey` instead of `GetPlatformPrivateKey` and move all the key related functionality to a separate file?
> 
> If you like I could throw in my implementation of exactly that and a bit of JSONAPI paging and filtering on top. But I also don't want to get in the way of what you already did @shreddedbacon .

I didn't know they were different. In that case it makes sense to name them a neutral name, in their own file. I'm happy to adjust my PR as I'm also currently using my branch to build out an operator for our platform.

How far along is your work on "Regular TLS" @scheying ?

phamann:
> Would it make sense to give the key-related functions neutral names like GetPrivateKey instead of GetPlatformPrivateKey and move all the key related functionality to a separate file?

@scheying This is a great question and in theory makes sense to me. I'm afraid that I'm not an internal SME (subject-matter-expert) on our TLS APIs, so will get someone on that team to have a look and share their thoughts here.

shreddedbacon:
> Thanks for the speedy changes to this, @shreddedbacon, I appreciate your patience here. It's looking much better! Most of my comments are minor comment style changes, some small copy/pasta errors and a question around using pointers for the pagination input structs.
> 
> My last major change request would be to update the tests to be more inline with the style we use in the package. Which would be: one `TestClient_PlatformPrivateKey` function which creates a key (defers its deletion) then uses that to perform the other operations (list, update, delete), with each operation assert the properties or length returned are valid. We also have seperate functions to test the validations. An example of this style can be found in some of the recent PRs, such as users: https://github.com/fastly/go-fastly/blob/master/fastly/user_test.go#L19-L133
> 
> Hope that helps. After this last change set, we should be good to merge!

No problem at all. Thanks for your patience.
I was mainly using existing code to try and find how things were styled, but I guess it has changed over time.

Will adjust again!

fastly/platform_tls.go
```diff
@@ -0,0 +1,380 @@
...
+// formatFilters converts user input into query parameters for filtering
+func (i *ListPrivateKeysInput) formatFilters() map[string]string {
+	result := map[string]string{}
+	pairings := map[string]interface{}{
+		"filter[in_use]": i.FilterInUse,
+		"page[size]":     i.PageSize,
+		"page[number]":   i.PageNumber,
+	}
+	// NOTE: This setup means we will not be able to send the zero value
+	// of any of these filters. It doesn't appear we would need to at present.
```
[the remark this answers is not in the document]
shreddedbacon: I'm happy to adjust this to using a pointer if it is the way forward.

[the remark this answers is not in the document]
shreddedbacon: Have you got an example PR for where this was implemented?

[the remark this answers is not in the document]
phamann: Sure: https://github.com/fastly/go-fastly/pull/142/files

shreddedbacon:
Hi again @phamann 
I've made adjustments but just want to check with you on something as I couldn't see it on any other Get requests.

For the filters to actually work, I had to add the header 
```
Headers: map[string]string{
  "Accept": "application/vnd.api+json",
}
```
As described in the API documentation https://docs.fastly.com/api/platform-tls#tls_bulk_certificates_81cc5acbf847f71ecd4068ed58bfc5c5

When I remove this header, the filter no longer works. Is there something I'm missing here? I noticed a few other functions use filters and have the same `Accept` heading requirement in the API, but yet they haven't had to define the header for this.

phamann:
> For the filters to actually work, I had to add the header. When I remove this header, the filter no longer works. Is there something I'm missing here?

@shreddedbacon You're not missing anything here. This is another nuance of our JSONAPI API requests and not your implementation. For those requests you are using the `client.Get()` helper which is intended for plain HTTP requests and has no knowledge about JSONAPI. It would be better if we had a `GetJSONAPI()` helper as per the other methods `PutJSONAPI` etc (which I've just realised we don't have...) and that would internally call [`RequestJSONAPI`](https://github.com/fastly/go-fastly/blob/master/fastly/client.go#L285-L305) which sets the correct `Accept` and `Content-Type` JSONAPI header values.

shreddedbacon:
Awesome. Thanks for smashing through this with me, appreciate it!

fastly/platform_tls.go
```diff
@@ -0,0 +1,380 @@
...
+	ID   string `jsonapi:"primary,tls_domain"`
+	Type string `jsonapi:"attr,type"`
+}
+
+// ListBulkCertificatesInput .
+type ListBulkCertificatesInput struct {
+	PageNumber int // The page index for pagination.
+	PageSize   int // The number of keys per page.
+	// `tls_domains.id` filter seems to work where `tls_domain.id` does not, documentation wrong?
+	// https://docs.fastly.com/api/platform-tls#tls_bulk_certificates_81cc5acbf847f71ecd4068ed58bfc5c5
```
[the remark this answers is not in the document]
thommahoney: Both filtering parameters should be supported, with a preference for the plural. I'll follow up with the appropriate teams.

fastly/platform_tls.go
```diff
@@ -0,0 +1,350 @@
...
+)
+
+// GetPrivateKeyInput is an input to the GetPrivateKey function.
+// Allowed values for the fields are described at https://docs.fastly.com/api/platform-tls.
+type GetPrivateKeyInput struct {
+	ID string
+}
+
+// PrivateKey represents a private key is used to sign a Certificate.
+type PrivateKey struct {
```
thommahoney, line 18: I'd be in favor of moving the `PrivateKey` structs and related requests into their own file. The APIs aren't specific to Platform TLS and are shared with the Fastly TLS product, as another reviewer noted.

fastly/platform_tls.go
```diff
@@ -0,0 +1,350 @@
...
+}
+
+// ListBulkCertificatesInput is used as input to the ListBulkCertificates function.
+type ListBulkCertificatesInput struct {
+	PageNumber *uint // The page index for pagination.
+	PageSize   *uint // The number of keys per page.
+	// `tls_domains.id` filter seems to work where `tls_domain.id` does not, documentation wrong?
+	// https://docs.fastly.com/api/platform-tls#tls_bulk_certificates_81cc5acbf847f71ecd4068ed58bfc5c5
+	FilterTLSDomainIDMatch  *string // Filter certificates by their matching, fully-qualified domain name. Returns all partial matches. Must provide a value longer than 3 characters.
+	FilterTLSDomainsIDMatch *string // Filter certificates by their matching, fully-qualified domain name. Returns all partial matches. Must provide a value longer than 3 characters.
```
thommahoney, line 185: While both of these are supported, I'd prefer to have only the plural one included in this library as we work to have people migrate away from the singular filter name. Could I ask that you remove `FilterTLSDomainIDMatch` throughout and keep `FilterTLSDomainsIDMatch`?

shreddedbacon:
Hi @thommahoney. 
Adjustments made 🤞

phamann:
Looking good! Once again thank you for your contribution :+1:

I'll let you know once released.

phamann:
This was released in [v1.10.0](https://github.com/fastly/go-fastly/releases/tag/v1.10.0) :tada:

PR #162, opened 2020-05-06
Add Rackspace Cloud Files endpoint support

ezkl:
See [the docs](https://docs.fastly.com/en/guides/log-streaming-cloudfiles) for more information on Cloud Files logging endpoints.

### DIFF
--- a/fastly/cloudfiles.go
+++ b/fastly/cloudfiles.go
@@ -1,0 +1,255 @@
+package fastly
+
+import (
+	"fmt"
+	"net/url"
+	"sort"
+	"time"
+)
+
+// Cloudfiles represents a Cloudfiles response from the Fastly API.
+type Cloudfiles struct {
+	ServiceID string `mapstructure:"service_id"`
+	Version   int    `mapstructure:"version"`
+
+	Name              string     `mapstructure:"name"`
+	User              string     `mapstructure:"user"`
+	AccessKey         string     `mapstructure:"access_key"`
+	BucketName        string     `mapstructure:"bucket_name"`
+	Path              string     `mapstructure:"path"`
+	Region            string     `mapstructure:"region"`
+	Placement         string     `mapstructure:"placement"`
+	Period            uint       `mapstructure:"period"`
+	GzipLevel         uint       `mapstructure:"gzip_level"`
+	Format            string     `mapstructure:"format"`
+	FormatVersion     uint       `mapstructure:"format_version"`
+	ResponseCondition string     `mapstructure:"response_condition"`
+	MessageType       string     `mapstructure:"message_type"`
+	TimestampFormat   string     `mapstructure:"timestamp_format"`
+	PublicKey         string     `mapstructure:"public_key"`
+	CreatedAt         *time.Time `mapstructure:"created_at"`
+	UpdatedAt         *time.Time `mapstructure:"updated_at"`
+	DeletedAt         *time.Time `mapstructure:"deleted_at"`
+}
+
+// cloudfilesByName is a sortable list of Cloudfiles.
+type cloudfilesByName []*Cloudfiles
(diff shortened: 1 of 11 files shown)


fastly/cloudfiles.go
```diff
@@ -0,0 +1,255 @@
...
+
+	Name              string `form:"name,omitempty"`
+	User              string `form:"user,omitempty"`
+	AccessKey         string `form:"access_key,omitempty"`
+	BucketName        string `form:"bucket_name,omitempty"`
+	Path              string `form:"path,omitempty"`
+	Region            string `form:"region,omitempty"`
+	Placement         string `form:"placement,omitempty"`
+	Period            uint   `form:"period,omitempty"`
+	GzipLevel         uint   `form:"gzip_level,omitempty"`
```
phamann, line 93: We should discuss how we want to handle scalar values and `omitempty` in the update input structs. We've recently started to move other update structs over to pointer values. The problem is the combination of Go's type zero values, JSON marshalling with the `omitempty` struct field tag, and a users intent to actually set a zero value. This blog post does a much better job at explaining the problem and well worth a read: https://willnorris.com/2014/05/go-rest-apis-and-pointers/ .

Therefore this current implementation would make it impossible to update these values to their zero values i.e. `0` in this case. A more in-depth explanation and discussion of recent decisions made in the library can be found in this PR: https://github.com/fastly/go-fastly/pull/142#discussion_r369001840

fastly/cloudfiles.go
```diff
@@ -0,0 +1,255 @@
...
+	if err != nil {
+		return err
+	}
+
+	var r *statusResp
+	if err := decodeJSON(&r, resp.Body); err != nil {
+		return err
+	}
+	if !r.Ok() {
+		return fmt.Errorf("Not Ok")
```
phamann, line 252: I can see we are using this in all of your other PRs too, maybe we should consider moving to an errors constant in [`errors.go`](https://github.com/fastly/go-fastly/blob/master/fastly/errors.go) and make the message more descriptive.

fastly/cloudfiles.go
```diff
@@ -0,0 +1,255 @@
...
+	Name string
+
+	NewName           string `form:"name,omitempty"`
+	AccessKey         string `form:"access_key,omitempty"`
+	BucketName        string `form:"bucket_name,omitempty"`
+	Path              string `form:"path,omitempty"`
+	Region            string `form:"region,omitempty"`
+	Placement         string `form:"placement,omitempty"`
+	Period            uint   `form:"period,omitempty"`
+	GzipLevel         uint   `form:"gzip_level,omitempty"`
```
phamann, line 180: We should discuss how we want to handle scalar values and omitempty in the update input structs. We've recently started to move other update structs over to pointer values. The problem is the combination of Go's type zero values, JSON marshalling with the omitempty struct field tag, and a users intent to actually set a zero value. This blog post does a much better job at explaining the problem and well worth a read: https://willnorris.com/2014/05/go-rest-apis-and-pointers/ .

Therefore this current implementation would make it impossible to update these values to their zero values i.e. 0 in this case. A more in-depth explanation and discussion of recent decisions made in the library can be found in this PR: https://github.com/fastly/go-fastly/pull/142#discussion_r368996195

ezkl: I believe I've addressed this in https://github.com/fastly/go-fastly/pull/162/commits/9c90c20aa2b31ea93689b69aa2f102dd24af9101.

fastly/cloudfiles.go
ezkl: I've addressed this in https://github.com/fastly/go-fastly/pull/162/commits/8fc2b9cbc39c3e0906b61233959df938d6d3dcfb.

PR #164, opened 2020-05-06
Add Elasticsearch endpoint support

ezkl:
See [the docs](https://docs.fastly.com/en/guides/log-streaming-elasticsearch) for more information on Elasticsearch logging endpoints.


fastly/elasticsearch_test.go
```diff
@@ -0,0 +1,342 @@
...
+}
+
+func TestClient_ListElasticsearch_validation(t *testing.T) {
+	var err error
+	_, err = testClient.ListElasticsearch(&ListElasticsearchInput{
+		Service: "",
+	})
+	if err != ErrMissingService {
+		t.Errorf("bad error: %s", err)
+	}
```
mccurdyc, line 242: I think this should also check that the Version is not zero.

https://github.com/fastly/go-fastly/pull/164/files#diff-d244653eb384ba4e0eb5a2753701f7a3R61

PR #165, opened 2020-05-06
Update FTP logging endpoint

ezkl:
* Rename Username to User for consistency with API field
* Add PublicKey
* Use pointers in CreateFTPInput and UpdateFTPInput structs and use helpers in tests
* Change GzipLevel from uint8 to uint for consistency


fastly/ftp.go
```diff
@@ -15,11 +15,12 @@ type FTP struct {
 	Name              string     `mapstructure:"name"`
 	Address           string     `mapstructure:"address"`
 	Port              uint       `mapstructure:"port"`
-	Username          string     `mapstructure:"user"`
+	User              string     `mapstructure:"user"`
```
mccurdyc, line 18: Yay! I picked up on this while working on the CLI.

fastly/ftp.go
```diff
@@ -15,11 +15,12 @@ type FTP struct {
...
 	Password          string     `mapstructure:"password"`
+	PublicKey         string     `mapstructure:"public_key"`
 	Path              string     `mapstructure:"path"`
 	Period            uint       `mapstructure:"period"`
-	GzipLevel         uint8      `mapstructure:"gzip_level"`
+	GzipLevel         uint       `mapstructure:"gzip_level"`
```
mccurdyc, line 23: Hm. So `uint8` makes sense for this field.

* [S3 uses `uint`](https://github.com/fastly/go-fastly/blob/d5f2d007fd1adeb514b6efd8927913fff21be6b4/fastly/s3.go#L32)
* [GCS uses `uint8`](https://github.com/fastly/go-fastly/blob/d5f2d007fd1adeb514b6efd8927913fff21be6b4/fastly/gcs.go#L21)

mccurdyc: @phamann What's your thought on this?

mccurdyc:
@phamann and I talked about this. While this is what we would like the API to be eventually, we can't make some of these changes right now without releasing a major version of the client, which we aren't ready to do.

So, for this change, we will focus on additions to the API, but no in-place changes and will note for these changes to be made in a v2 release of the client.

fastly/ftp.go
```diff
@@ -244,7 +247,7 @@ func (c *Client) DeleteFTP(i *DeleteFTPInput) error {
 		return err
 	}
 	if !r.Ok() {
-		return fmt.Errorf("Not Ok")
+		return ErrStatusNotOk
```
mccurdyc, line 250: Unfortunately, this is a breaking change.

fastly/ftp.go
mccurdyc: Unfortunately, we can't make this change because it would break backward compatibility.

fastly/ftp.go
mccurdyc, line 18: Unfortunately, we can't make this change because it would break backward compatibility.

fastly/ftp.go
```diff
@@ -80,19 +81,20 @@ type CreateFTPInput struct {
...
+	Password          *string `form:"password,omitempty"`
+	PublicKey         *string `form:"public_key,omitempty"`
+	Path              *string `form:"path,omitempty"`
+	Period            *uint   `form:"period,omitempty"`
+	FormatVersion     *uint   `form:"format_version,omitempty"`
+	GzipLevel         *uint   `form:"gzip_level,omitempty"`
+	Format            *string `form:"format,omitempty"`
+	ResponseCondition *string `form:"response_condition,omitempty"`
+	TimestampFormat   *string `form:"timestamp_format,omitempty"`
+	Placement         *string `form:"placement,omitempty"`
```
mccurdyc, line 97: Unfortunately, we can't make this change because it would break backward compatibility.

fastly/ftp.go
```diff
@@ -166,19 +168,20 @@ type UpdateFTPInput struct {
...
+	User              *string `form:"user,omitempty"`
+	Password          *string `form:"password,omitempty"`
+	Path              *string `form:"path,omitempty"`
+	Period            *uint   `form:"period,omitempty"`
+	FormatVersion     *uint   `form:"format_version,omitempty"`
+	GzipLevel         *uint   `form:"gzip_level,omitempty"`
+	Format            *string `form:"format,omitempty"`
+	ResponseCondition *string `form:"response_condition,omitempty"`
+	TimestampFormat   *string `form:"timestamp_format,omitempty"`
+	Placement         *string `form:"placement,omitempty"`
```
mccurdyc, line 184: Unfortunately, we can't make this change because it would break backward compatibility.

fastly/fixtures/ftps/get.yaml
```diff
@@ -1,30 +1,32 @@
 ---
 version: 1
-rwmutex: {}
 interactions:
 - request:
     body: ""
     form: {}
     headers:
-      Fastly-Key:
```
mccurdyc, line 9: @ezkl @phamann not sure why this got removed.

fastly/ftp_test.go
```diff
@@ -169,6 +176,14 @@ func TestClient_FTPs(t *testing.T) {
 	if uftp.Name != "new-test-ftp" {
 		t.Errorf("bad name: %q", uftp.Name)
 	}
+	// TODO (v2): This is a bug where updates to zero-values are omitted due to the
```
mccurdyc, line 179: @phamann test that exercises the bug with a note for v2.

fastly/fixtures/ftps/get.yaml
phamann: We now have a filter which removes all `Fastly-Key` headers from VCR recordings. See: https://github.com/fastly/go-fastly/blob/master/fastly/fastly_test.go#L61-L65 

These files would have been produced before that change.

fastly/ftp.go
```diff
@@ -166,19 +168,20 @@ type UpdateFTPInput struct {
...
-	FormatVersion     uint   `form:"format_version,omitempty"`
-	GzipLevel         uint8  `form:"gzip_level,omitempty"`
-	Format            string `form:"format,omitempty"`
-	ResponseCondition string `form:"response_condition,omitempty"`
-	TimestampFormat   string `form:"timestamp_format,omitempty"`
-	Placement         string `form:"placement,omitempty"`
+	NewName           string  `form:"name,omitempty"`
+	Address           string  `form:"address,omitempty"`
+	Port              uint    `form:"port,omitempty"`
+	PublicKey         *string `form:"public_key,omitempty"`
```
phamann, line 174: I'd prefer it if the structs interface was consistent, so don't think we should introduce the pointer pattern, even for a new field.

PR #172, opened 2020-05-06
Add OpenStack endpoint support

ezkl:
See [the docs](https://docs.fastly.com/en/guides/log-streaming-openstack) for more information on OpenStack logging endpoints.


fastly/openstack.go
```diff
@@ -0,0 +1,255 @@
...
+	// Service is the ID of the service. Version is the specific configuration
+	// version. Both fields are required.
+	Service string
+	Version int
+
+	// Name is the name of the Openstack to update.
+	Name string
+
+	NewName           *string `form:"name,omitempty"`
+	AccessKey         *string `form:"access_key,omitempty"`
```
mccurdyc, line 174: I'm going to dig deeper, but why isn't `User` here?

mccurdyc: I think this may have just been a typo. I don't see a reason for it to be missing.

mccurdyc: I just tried updating the field on manage.fastly.com and it is editable on updates. So I think it should be added.

PR #176, opened 2020-05-06
Add raw access to stats JSON responses

pteichman:
This PR adds GetStatsJSON and GetRealtimeStatsJSON, which give the caller control over the JSON objects decoded from stats responses. The existing GetStats and GetRealtimeStats (unchanged) discard the original structure.

### DIFF
--- a/fastly/client.go
+++ b/fastly/client.go
@@ -342,8 +342,9 @@ func checkResp(resp *http.Response, err error) (*http.Response, error) {
 	}
 }
 
-// decodeJSON is used to decode an HTTP response body into an interface as JSON.
-func decodeJSON(out interface{}, body io.ReadCloser) error {
+// decodeBodyMap is used to decode an HTTP response body into a mapstructure struct.
+// It closes `body`.
+func decodeBodyMap(out interface{}, body io.ReadCloser) error {
 	defer body.Close()
 
 	var parsed interface{}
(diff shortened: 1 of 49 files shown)


fastly/client.go
```diff
@@ -342,8 +342,9 @@ func checkResp(resp *http.Response, err error) (*http.Response, error) {
 	}
 }
 
-// decodeJSON is used to decode an HTTP response body into an interface as JSON.
-func decodeJSON(out interface{}, body io.ReadCloser) error {
+// decodeBodyMap is used to decode an HTTP response body into a mapstructure struct.
+// It closes `body`.
+func decodeBodyMap(out interface{}, body io.ReadCloser) error {
```
phamann, line 347: I know that the signature of this function was already like this, however it seems more natural to me (and more closely follwos other Go convention `Unmarshal()` etc) if the arguments were flipped:

```suggestion
func decodeBodyMap(body io.ReadCloser, out interface{}) error {
```

The same applies to the `decodeMap` on line 361

pteichman: Heh, my new one in decodeMap was from standard library convention too: `copy(dst, src)` where the easy way to remember is it's the same order as a variable assignment.

It's an easy argument to say decodeBodyMap should match Go's json library though, and I'll make decodeMap match that so at least locally there's less confusion.

PR #179, opened 2020-05-11
Support for the link to the rule set deploy status

philippschulte:
This PR is necessary in order to be able to use the link of the rule set deploy status in [waflyctl](https://github.com/fastly/waflyctl).

The request methods for [GET/service/service_id/wafs/waf_id/ruleset](https://docs.fastly.com/api/waf#waf_ruleset_5e37b65346b88d394dcf3ba6b07e35d4) and [PATCH/service/service_id/wafs/waf_id/ruleset](https://docs.fastly.com/api/waf#waf_ruleset_3c076195f3616cf75dc9cf274916a768) share the same response struct. However, the responses of these endpoints are actually differnet. The `PATCH` endpoint doesn't support the `vcl` and the `last_push` fields.

I am just adding the `Link` field to the shared `Ruleset` response struct. This field will be empty and useless for the [GET/service/service_id/wafs/waf_id/ruleset](https://docs.fastly.com/api/waf#waf_ruleset_5e37b65346b88d394dcf3ba6b07e35d4) request function but I don't think this matters because the response struct for [PATCH/service/service_id/wafs/waf_id/ruleset](https://docs.fastly.com/api/waf#waf_ruleset_3c076195f3616cf75dc9cf274916a768) also includes empty and useless fields as well. The `jsonapi` package doesn't allow me to decode the `Link` field and therefore I am using the standard library for the job.

I understand that this is not the nicest implementation but it doesn't introduce any breaking changes.

### DIFF
--- a/fastly/fixtures/wafs/cleanup.yaml
+++ b/fastly/fixtures/wafs/cleanup.yaml
@@ -1,24 +1,22 @@
 ---
 version: 1
-rwmutex: {}
 interactions:
 - request:
     body: ""
     form: {}
     headers:
-      Fastly-Key:
-      - xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
       User-Agent:
-      - FastlyGo/0.4.3.dev (+github.com/fastly/go-fastly; go1.12)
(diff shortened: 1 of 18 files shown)


philippschulte:
@phamann I have added the missing tests. Thank you!

PR #183, opened 2020-05-12
Fix staticcheck issues

dgryski:
Note that these are almost all style issues, except for the issue
in request_setting_test.go where the same field was on the left
and right hand sides of an inequality check.


mccurdyc:
Thank you! Closing because of https://github.com/fastly/go-fastly/pull/184/commits/79b373bdb04c0db69d8169edb8a179e5b3e1dad8.

PR #184, opened 2020-05-12
ci: add GH Action for fmt, vet, staticcheck, test

mccurdyc:
## Proposed Change(s)
* Adds a GitHub Action CI workflow
* CI workflow with fmt, vet, staticcheck and test
* Adds necessary make targets
* Address staticcheck errors
   * https://github.com/fastly/go-fastly/pull/184/commits/79b373bdb04c0db69d8169edb8a179e5b3e1dad8
   * There was a bug in the test where a value was erroneously compared to itself instead of an updated value.
* Adds `make help` menu.

```bash
$ make help
all                            Runs all of the required cleaning and verification targets.
changelog                      Generates the full project changelog.
* check-fmt                      A check which lists improperly-formatted files, if they exist.
* check-imports                  A check which lists improperly-formatted imports, if they exist.
* check-mod                      A check which lists extraneous dependencies, if they exist.
* dev-dependencies               Downloads the necessesary dev dependencies.
fix-fixtures                   Updates test fixtures with a specified default service ID.
* fiximports                     Properly formats and orders imports.
* fmt                            Properly formats Go files and orders dependencies.
help                           Prints this help menu.
* mod-download                   Downloads the Go module.
release-changelog              Generates the changelog for a specific release.
* staticcheck                    Runs the staticcheck linter.
test-full                      Runs the tests with VCR disabled (i.e., makes external calls).
test-race                      Runs the test suite with the -race flag to identify race conditions, if they exist.
test                           Runs the test suite with VCR mocks enabled.
* tidy                           Cleans the Go module.
* vet                            Identifies common errors.
```

`*` indicates that this target was added in this PR.

## Relevant Link(s)

* Inspiration came from [`fastly/cli`](https://github.com/fastly/cli/blob/master/.github/workflows/pr_test.yml)

## Validation
```
$ make
```

```
git clone git@github.com:mccurdyc/go-fastly.git /tmp/go-fastly
(
    cd /tmp/go-fastly
    git checkout mccurdyc/ci-checks
    make
    make help
    make check-fmt
    make check-mod
    make check-imports
    rm -rf /tmp/go-fastly
)
```


Makefile
```diff
@@ -91,25 +92,44 @@ linux:
...
 # update fixtures default service ID
 fix-fixtures:
 	@echo "==> Updating fixtures"
 	@$(CURRENT_DIR)/scripts/fixFixtures.sh ${FASTLY_TEST_SERVICE_ID}
 .PHONY: fix-fixtures
+
+# lists improperly-formatted files if they exist
+check-fmt:
+	@gofmt -s -l ${GOFILES}
+	@printf "\nPlease, run 'make fmt'.\n"
```
mccurdyc, line 121: This needs to check for a non-zero exit code from the above command before printing this.

mccurdyc:
> - [ ] We should also install golint and add an associated make lint target.
I didn't do this because of this issue - https://github.com/golang/go/issues/38968

> - [x] We are no longer using dep so we should remove the default bootstrap make target and replace with a make all which runs all of the linting steps and tests as the default target

> - [x] We should run all of the above commands and commit any fixes as part of this PR

> - [x] We should document the tooling requirements in the README.md somehow

> - [x] As this replaces the Travis build, we should remove the .travis.yml

> I will follow up in a separate PR with some of our other Fastly OSS standards, such as a SECURITY.md etc

:+1:

README.md
```diff
@@ -165,7 +182,7 @@ Example (`make test`, `make fix-fixtures`):
 export FASTLY_TEST_SERVICE_ID="SERVICEID"
 export FASTLY_API_KEY="TESTAPIKEY"
 
-make test TESTARGS="-run=NewClient"
```
mccurdyc, line 168: `NewClient` doesn't exist and through me off for a while when I was trying to test.

fastly/errors.go
```diff
@@ -11,147 +11,150 @@ import (
...
-var ErrMissingIntermediatesBlob = errors.New("Missing required field 'IntermediatesBlob'")
+var ErrMissingIntermediatesBlob = errors.New("missing required field 'IntermediatesBlob'")
 
 // ErrStatusNotOk is an error that indicates that indicates that the response body returned
 // by the Fastly API was not `{"status": "ok"}`
-var ErrStatusNotOk = errors.New("Unexpected 'status' field in API response body")
+var ErrStatusNotOk = errors.New("unexpected 'status' field in API response body")
+
+// ErrNotOK is a generic error indicating that something is not okay.
+var ErrNotOK = errors.New("not ok")
```
mccurdyc, line 157: This is new.

fastly/event_logs_test.go
```diff
@@ -6,8 +6,6 @@ import (
 	"testing"
 )
 
-var testEventID = "3OMewexIMbzrQj77xxxxxx"
```
mccurdyc, line 9: This was dead.

fastly/request_setting_test.go
```diff
@@ -118,7 +118,7 @@ func TestClient_RequestSettings(t *testing.T) {
 	if rs.ForceMiss != nrs.ForceMiss {
 		t.Errorf("bad force_miss: %t (%t)", rs.ForceMiss, nrs.ForceMiss)
 	}
-	if rs.ForceSSL != rs.ForceSSL {
+	if rs.ForceSSL != nrs.ForceSSL {
```
mccurdyc, line 121: BUG.

tools/tools.go
```diff
@@ -0,0 +1,13 @@
+// +build tools
+
+// This follows the agreed-upon current best approach for adding developer tooling to your Module.
+//
+// For more details, refer to this section in the Go Wiki:
+// https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module
```
mccurdyc, line 6: See above for more information on the chosen approach.

fastly/errors.go
```diff
@@ -11,147 +11,150 @@ import (
...
 
 // ErrMissingName is an error that is returned when an input struct requires a
 // "Name" key, but one was not set.
-var ErrMissingName = errors.New("Missing required field 'Name'")
+var ErrMissingName = errors.New("missing required field 'Name'")
 
 // ErrMissingKey is an error that is returned when an input struct requires a
 // "Key" key, but one was not set.
-var ErrMissingKey = errors.New("Missing required field 'Key'")
+var ErrMissingKey = errors.New("missing required field 'Key'")
```
mccurdyc, line 42: Do note that these are part of the exported API. I assume that it is acceptable to change the error value, right?

.github/workflows/pr_test.yml
```diff
@@ -0,0 +1,76 @@
+on: pull_request
+name: Test
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v1
```
phamann, line 8: See setup-go comments for reasoning

```suggestion
      uses: actions/checkout@v2
```

.github/workflows/pr_test.yml
```diff
@@ -0,0 +1,76 @@
...
+    - name: Install Go
+      uses: actions/setup-go@v1
```
phamann, line 10: V2 has now been released and fixes the GOPATH issue we had a workaround for in the CLI, which means we can also remove the shell step below. 

```suggestion
      uses: actions/setup-go@v2
```

.github/workflows/pr_test.yml
```diff
@@ -0,0 +1,76 @@
...
+      with:
+        go-version: 1.14.x
+    - name: Set GOPATH
+      # Temporary fix, see: https://github.com/actions/setup-go/issues/14
+      run: |
+        echo "::set-env name=GOPATH::$(go env GOPATH)"
+        echo "::add-path::$(go env GOPATH)/bin"
+      shell: bash
```
phamann, line 18: Had a check back in the linked issue and seems this hack has now been resolved in the v2 of the action. Therefore, we can remove and bump checkout and go-install to v2.

```suggestion
```

.github/workflows/pr_test.yml
```diff
@@ -0,0 +1,76 @@
...
+    - name: Restore cache
+      id: cache
+      uses: actions/cache@v1
```
phamann, line 21: This is now on v2.

```suggestion
      uses: actions/cache@v2
```

.github/workflows/pr_test.yml
```diff
@@ -0,0 +1,76 @@
...
+      with:
+        path: ~/go/bin
+        key: ${{ runner.os }}-go-bin-${{ hashFiles('**/go.sum') }}
+        restore-keys: |
+          ${{ runner.os }}-go-bin-
+    - name: Restore cache
+      uses: actions/cache@v1
```
phamann, line 28: As above:

```suggestion
      uses: actions/cache@v2
```

.github/workflows/pr_test.yml
```diff
@@ -0,0 +1,76 @@
...
+      run: make vet
+      shell: bash
+    - name: Staticcheck
+      run: make staticcheck
+      shell: bash
+  test:
+    strategy:
+      matrix:
+        go-version: [1.14.x]
+        platform: [ubuntu-latest, macos-latest, windows-latest]
```
phamann, line 53: I'm personally happy for us to only test this project on a linux architecture (which is what we currently do in Travis), we use all three in the CLI project as we distribute a binary across all architectures. However as this is a package that gets consumed by others I don't think its necessary and slows down PR builds considerably. 
```suggestion
        platform: [ubuntu-latest]
```

.github/workflows/pr_test.yml
```diff
@@ -0,0 +1,76 @@
...
+    runs-on: ${{ matrix.platform }}
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v1
```
phamann, line 57:
```suggestion
      uses: actions/checkout@v2
```

.github/workflows/pr_test.yml
```diff
@@ -0,0 +1,76 @@
...
+    - name: Install Go
+      uses: actions/setup-go@v1
```
phamann, line 59:
```suggestion
      uses: actions/setup-go@v2
```

.github/workflows/pr_test.yml
```diff
@@ -0,0 +1,76 @@
...
+      with:
+        go-version: ${{ matrix.go-version }}
+    - name: Set GOPATH
+      # Temporary fix, see: https://github.com/actions/setup-go/issues/14
+      run: |
+        echo "::set-env name=GOPATH::$(go env GOPATH)"
+        echo "::add-path::$(go env GOPATH)/bin"
```
phamann, line 66: See above for reasoning. 

```suggestion
```

.github/workflows/pr_test.yml
```diff
@@ -0,0 +1,76 @@
...
+    - name: Restore cache
+      uses: actions/cache@v1
```
phamann, line 68:
```suggestion
      uses: actions/cache@v2
```

.travis.yml
```diff
@@ -1,18 +0,0 @@
-dist: trusty
```
phamann, line 1: Maybe we should first validate the new GitHub Action flow on a couple of PRs before deleting this?

README.md
```diff
@@ -1,24 +1,24 @@
 Go Fastly
 =========
-[![Build Status](http://img.shields.io/travis/fastly/go-fastly.svg?style=flat-square)][travis]
 [![Go Documentation](http://img.shields.io/badge/go-documentation-blue.svg?style=flat-square)][godocs]
 
-[travis]: http://travis-ci.org/fastly/go-fastly
-[godocs]: http://godoc.org/github.com/fastly/go-fastly
+[godocs]: https://pkg.go.dev/github.com/fastly/go-fastly?tab=overview
```
phamann, line 5: 🔥

README.md
```diff
@@ -125,6 +125,23 @@ fmt.Printf("%t\n", activeVersion.Locked)
 More information can be found in the
 [Fastly Godoc](https://godoc.org/github.com/fastly/go-fastly).
 
+Developing
+-------
+
+1. Clone the project to your preferred directory, using your preferred method
+2. Download the module and accompanying developer tooling
```
phamann, line 132:
```suggestion
1. Clone the project to your preferred directory, using your preferred method.
2. Download the module and accompanying developer tooling.
```

fastly/errors.go
phamann: I personally think this is acceptable and don't think it constitutes a breaking change as consumers _should_ be asserting against the types token and not the value.

fastly/errors.go
phamann: I'd agree that `ErrNotOK` is the correct variant. And would prefer to go with that.

fastly/request_setting_test.go
phamann: Nice!

scripts/check-imports.sh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/env bash
+# https://github.com/hashicorp/vault/blob/master/scripts/gofmtcheck.sh
```
phamann, line 2: Copy / paste error?

Makefile
```diff
@@ -1,123 +1,93 @@
...
-.PHONY: dev
-
-# linux builds the linux binary
-linux:
-	@env \
-		GOOS="linux" \
-		GOARCH="amd64" \
-		$(MAKE) -f "${MKFILE_PATH}" build
-.PHONY: linux
+all: tidy fmt fiximports test vet staticcheck ## Runs all of the required cleaning and verification targets.
```
phamann, line 21: Maybe we should add a target to install the tooling required to run these tasks, some people may not have `goimport` and `staticcheck` installed.

mccurdyc: So `tidy` would install these tools, right, because they are part of the Go module and `go` commands would pull down dependencies.

I may be wrong. Maybe a `download` target that does a `go mod download`.

mccurdyc: I was wrong.

```bash
$ rm -rf $(which staticcheck)
$ make
```

Fails with:

```bash
==> Running staticcheck
/bin/bash: staticcheck: command not found
make: *** [Makefile:80: staticcheck] Error 127
```

Adding the `download` target.

mccurdyc: Ahh I was still wrong. This would just allow versioning of the tools, but doesn't actually install the binaries. So, yes, you were correct, we would also need an "install" target. I'm going to call it `dev-dependencies`.

mccurdyc: Confirmed that it worked:

```bash
$ rm -rf $(which goimports)
$ rm -rf $(which staticcheck)
$ make
==> Downloading Go module
==> Downloading development dependencies
==> Tidying module
==> Running gofmt
==> Fixing imports
==> Testing go-fastly
ok      github.com/fastly/go-fastly/fastly      (cached)
==> Running go vet
==> Running staticcheck
make  7.83s user 0.82s system 208% cpu 4.154 total
```

Makefile
```diff
@@ -1,123 +1,104 @@
...
+mod-download: ## Downloads the Go module.
+	@echo "==> Downloading Go module"
+	@go mod download
+.PHONY: mod-download
+
+dev-dependencies: ## Downloads the necessesary dev dependencies.
+	@echo "==> Downloading development dependencies"
+	@go install honnef.co/go/tools/cmd/staticcheck
+	@go install golang.org/x/tools/cmd/goimports
+.PHONY: dev-dependencies
```
phamann, line 38: Doesn't this dirty the `go.mod` `go.sum` manifests?

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+  - "1.11.x"
+  - "1.12.x"
+  - "1.13.x"
+  - "1.14.x"
+
+branches:
+  only:
+    - master
+
+script: make build test test-race
```
phamann: Looks like this needs to be updated to use the new make targets.

Makefile
mccurdyc: Yes, it does add to the Module manifest files. But I wouldn't consider this "dirtying" them. So the goal of adding them to the module is to ensure that developers are using the same version of the dependency.

For example, you wouldn't want someone running a super old version of staticcheck that doesn't yet include new, valuable, checks.

mccurdyc: Someone shared https://github.com/bwplotka/bingo internally and I did look into it and it does solve shortcomings of the current advice of the community (i.e., `package tools` approach), namely:

1. Avoids dependency and tools defining different constraints on transitive dependencies.
2. Tools aren't included as part of the `go-fastly` module i.e., what package consumers are interested in.

I don't have strong opinions, but in my opinion, it feels like a hacky solution built on top of an already hacky solution. Unless we are concerned with package consumers receiving tool dependencies and until we face point \#1, described above, I don't think the added complexity of another tool and managing sub-modules is worth it.

Makefile
```diff
@@ -1,123 +1,104 @@
...
-.PHONY: dev
-
-# linux builds the linux binary
-linux:
-	@env \
-		GOOS="linux" \
-		GOARCH="amd64" \
-		$(MAKE) -f "${MKFILE_PATH}" build
-.PHONY: linux
+all: mod-download dev-dependencies tidy fmt fiximports test vet staticcheck ## Runs all of the required cleaning and verification targets.
```
mccurdyc, line 21: @phamann Should this run `test-race` instead? Or in addition?

mccurdyc: Unless we use locks, channels, goroutines, `-race` probably isn't going to identify anything and with our `test-race` make target there is an additional overhead of not running the tests in parallel.

mccurdyc: I do see that [we use a single mutex in a Request to the API](https://github.com/fastly/go-fastly/blob/9f19db2dee031b5db530fdfef9415dfb9ed02d41/fastly/client.go#L253-L254), but that's the only concurrency primitive in this project.

mccurdyc: My logic is a bit flawed. While probably unlikely, we could be making a call to a function that uses concurrency primitives and does so in an unsafe manner. 

It does protect against potential races being introduced. Therefore, in order to keep the iterative process of development as streamlined as possible, i'll just have `test-race` run in CI, not locally.

mccurdyc:
> - [x] Some improvements to the actions workflow yaml
- [x] Should consider adding a step to install the tooling deps in a make target
   - Some discussion above on this topic.
- [x] Maybe we should wait before deleting the travis.yml file

mccurdyc:
@phamann I'm +1 for removing the `vendor` directory.

The value has diminished and having to remember to vendor new deps/versions of deps, as well as the diffs that come with it, often make reviewing PRs significantly more difficult.

TL;DR - the value add of the `vendor` directory is outweighed by the cost; removing.

PR #187, opened 2020-05-18
Add NullString helper

ezkl:
Adds a `NullString` helper that either returns a pointer to the string value passed in or nil if it the string is empty (`""`).

### DIFF
--- a/fastly/fastly.go
+++ b/fastly/fastly.go
@@ -70,6 +70,15 @@ func String(v string) *string {
 	return &v
 }
 
+// NullString is a helper that returns a pointer to the string value passed in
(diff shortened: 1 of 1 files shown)


fastly/fastly.go
```diff
@@ -70,6 +70,14 @@ func String(v string) *string {
 	return &v
 }
 
+// NullString is a helper that returns a pointer to string value passed in or nil if the string is empty.
```
phamann, line 73:
```suggestion
// NullString is a helper that returns a pointer to string value passed in or 
// nil if the string is empty.
```

PR #189, opened 2020-05-21
Get Public IPs (IPv6)

wbautistaEXPEDIA:
Exposed IPsV6 and AllIPs APIs. Added tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,6 @@ bin/
 pkg/
 
 .envrc
+
+# IDE specs
+*.iml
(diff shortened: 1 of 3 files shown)


wbautistaEXPEDIA:
> Why not leaving the `IPs` function as it is and just introducing a new function for retrieving the IPv6 addresses instead?

That's what I did. I think the diff in the PR may be adding to confusion, but it essentially leaves the original API as is, and exposes two new APIs (one for retrieving both sets, and one for the ipv6 set). See https://github.com/fastly/go-fastly/pull/189/files#diff-063a6e698147141c9ccad8e9632e0e13R22

wbautistaEXPEDIA:
@philippschulte 
Is there an ETA on merges and releases?

philippschulte:
@wbautistaEXPEDIA [v1.14.0](https://github.com/fastly/go-fastly/releases/tag/v1.14.0) has been released. Thanks again for your contribution!

PR #198, opened 2020-06-03
Add PublicKey field to S3 Updates

mccurdyc:


fastly/fastly_test.go
```diff
@@ -290,6 +290,68 @@ wMfrTEOvx0NxUM3rpaCgEmuWbB1G1Hu371oyr4srrr+N
 `
 }
 
+// pgpPublicKeyUpdate returns a PEM encoded PGP public key suitable for testing
+// updates to the public key field.
+func pgpPublicKeyUpdate() string {
```
mccurdyc, line 295: I'm "iffy" on doing this.

fastly/fixtures/s3s/create.yaml
```diff
@@ -30,6 +30,38 @@ interactions:
       - "12"
       placement:
       - waf_debug
+      public_key:
+      - |
```
mccurdyc, line 34: Public key existed for creates, just wasn't being tested.

fastly/fixtures/s3s/update.yaml
```diff
@@ -2,25 +2,85 @@
...
       Name:
       - test-s3
       Service:
       - 7i6HN3TK9wS159v2gPAZ8A
       Version:
-      - "6"
+      - "18"
       name:
       - new-test-s3
+      public_key:
```
mccurdyc, line 15: This is the new field

fastly/s3_test.go
```diff
@@ -27,10 +27,12 @@ func TestClient_S3s(t *testing.T) {
 			GzipLevel:                    9,
 			Format:                       "format",
 			FormatVersion:                2,
+			ResponseCondition:            "",
```
mccurdyc, line 30: Explicitly setting zero value just to show that it is covered.

However without adding additional setup --- i.e., creating a valid response condition --- we can't set a value here.

Maybe a comment?

fastly/s3_test.go
```diff
@@ -183,10 +197,11 @@ func TestClient_S3s(t *testing.T) {
...
 		us3, err = c.UpdateS3(&UpdateS3Input{
-			Service: testServiceID,
-			Version: tv.Number,
-			Name:    "test-s3",
-			NewName: "new-test-s3",
+			Service:   testServiceID,
+			Version:   tv.Number,
+			Name:      "test-s3",
+			NewName:   "new-test-s3",
+			PublicKey: pgpPublicKeyUpdate(),
```
mccurdyc, line 204: update public key

fastly/fastly_test.go
ezkl: Since there are fake keys all over the test suite, this doesn't bother me _too_ much. Worth adding to the list of testing cleanup TODOs, though.

fastly/s3_test.go
ezkl: The coverage here seems fine to me. If you think a comment will help highlight your concern in the future, go for it.

fastly/fastly_test.go
phamann: What are your concerns?

PR #203, opened 2020-06-23
Add support for Compute@Edge Packages

guyrichardson:
This update adds Package functionality to support the release of Wasm service resources.

In addition there are some Go formatting changes.


fastly/client.go
```diff
@@ -280,6 +287,46 @@ func (c *Client) RequestForm(verb, p string, i interface{}, ro *RequestOptions)
...
 }
 
+// RequestForm makes an HTTP request with the given interface being encoded as
+// form data.
+func (c *Client) RequestFormFile(verb, p string, i interface{}, f string, ro *RequestOptions) (*http.Response, error) {
+	file, err := os.Open(filepath.Clean(f))
+	if err != nil {
+		return nil, fmt.Errorf("error reading package: %v", err)
+	}
+	defer file.Close() // #nosec G307
```
mccurdyc, line 297: https://github.com/securego/gosec#available-rules

> G307: Deferring a method which returns an error

:+1:

fastly/fixtures/wasm_package/get.yaml
```diff
@@ -0,0 +1,48 @@
+---
+version: 1
+interactions:
+- request:
+    body: ""
+    form: {}
+    headers:
+      User-Agent:
+      - FastlyGo/1.15.0 (+github.com/fastly/go-fastly; go1.14.3)
+    url: https://api.fastly.com/service/2Q0B1WlDbii9DItrMCvIGL/version/2/package
```
mccurdyc, line 10: I'm not sure how we want to handle this.

For VCL services, we have this `fixFixtures` script that sets some default serviceID.

https://github.com/fastly/go-fastly/blob/master/scripts/fixFixtures.sh

fastly/fixtures/wasm_package/get.yaml
```diff
@@ -0,0 +1,48 @@
...
+    method: GET
+  response:
+    body: '{"version":2,"deleted_at":null,"updated_at":"2020-06-23T11:11:37Z","created_at":"2020-06-23T11:11:34Z","id":"0eh4MAGwaPjZerrjj44ZI6","metadata":{"name":"wasm-test","description":"Default
+      package template used by the Fastly CLI for Rust-based Compute@Edge projects.","authors":["guy.richardson+fastly@opencredo.com"],"language":"rust","size":2179443,"hashsum":"119a5ac6ec8a6dc29107b87de3bba5d98b6e89827310dd01f4e8cbf89234125bf4f7b81603ac7df42ce0205d047ac2caa53ec0cc52839112aed786cb7f27ac92"},"service_id":"2Q0B1WlDbii9DItrMCvIGL"}'
```
mccurdyc, line 14: Not sure if you want your email in there. You could replace with some fake address.

fastly/wasm.go
```diff
@@ -0,0 +1,96 @@
+package fastly
+
+import (
+	"fmt"
+	"io"
+	"time"
+)
+
+type WASMPackageMetadata struct {
```
mccurdyc, line 9: Just noting what @phamann pointed out in another place

https://webassembly.github.io/spec/core/intro/introduction.html?highlight=wasm#id1
```suggestion
type WasmPackageMetadata struct {
```

guyrichardson: This should be resolved.

fastly/fixtures/wasm_package/get.yaml
guyrichardson: Could you provide me with an appropriate default Wasm serviceID (at a guess a Wasm service in a fastly account?)... then I'll be able to update fixFixtures to replace both the Wasm and VCL serviceID.

fastly/fixtures/wasm_package/get.yaml
```diff
@@ -0,0 +1,48 @@
...
+    body: ""
+    form: {}
+    headers:
+      User-Agent:
+      - FastlyGo/1.15.0 (+github.com/fastly/go-fastly; go1.14.3)
+    url: https://api.fastly.com/service/6dV8eJUF4ViYvaSGIoEIFE/version/2/package
+    method: GET
+  response:
+    body: '{"metadata":{"name":"wasm-test","description":"Default package template
+      used by the Fastly CLI for Rust-based Compute@Edge projects.","authors":["guy.richardson+fastly@opencredo.com"],"language":"rust","size":2179443,"hashsum":"119a5ac6ec8a6dc29107b87de3bba5d98b6e89827310dd01f4e8cbf89234125bf4f7b81603ac7df42ce0205d047ac2caa53ec0cc52839112aed786cb7f27ac92"},"updated_at":"2020-06-24T08:11:45Z","created_at":"2020-06-24T08:11:44Z","version":2,"service_id":"6dV8eJUF4ViYvaSGIoEIFE","deleted_at":null,"id":"1VWqouO9TkE2B9fhKcPlBk"}'
```
phamann: Remove / change email address 😉

fastly/wasm.go
```diff
@@ -0,0 +1,96 @@
...
+	"io"
+	"time"
+)
+
+type WasmPackageMetadata struct {
+	Name        string
+	Description string
+	Authors     []string
+	Language    string
+	Size        int
```
phamann, line 14: We should consider making this an `int64` to be future proof. I can check internally.

fastly/wasm.go
```diff
@@ -0,0 +1,96 @@
...
+	HashSum     string
+}
+
+type WasmPackage struct {
+	ID        string
+	ServiceID string `mapstructure:"service_id"`
+	Version   int
+	CreatedAt *time.Time `mapstructure:"created_at"`
+	UpdatedAt *time.Time `mapstructure:"updated_at"`
+	DeletedAt *time.Time `mapstructure:"deleted_at"`
```
phamann, line 24: Please put the time based metadata fields last in the struct definition.

fastly/wasm.go
```diff
@@ -0,0 +1,96 @@
...
+// GetWasmPackageInput is used as input to the GetWasmPackage function.
+type GetWasmPackageInput struct {
+	// Service is the ID of the service.
+	// Version is the specific configuration version.
+	// Both fields are required.
+	Service string `mapstructure:"service_id"`
+	Version int    `mapstructure:"version"`
+}
+
+// GetWasmPackage retrieves Wasm package information for the given service and version
```
phamann, line 37:
```suggestion
// GetWasmPackage retrieves Wasm package information for the given service and version.
```

fastly/fastly_test.go
```diff
@@ -22,6 +22,9 @@ var testServiceID = serviceIDForTest()
 // Default ID of the testing service.
 var defaultTestServiceID = "7i6HN3TK9wS159v2gPAZ8A"
 
+var ServiceTypeVCL = "vcl"
+var ServiceTypeWasm = "wasm"
```
phamann, line 26: These can be constants.

```suggestion
const ( 
    // ServiceTypeVCL is the type for VCL services.
    ServiceTypeVCL = "vcl"
    // ServiceTypeWasm is the type for Wasm services.
    ServiceTypeWasm = "wasm"
)
```

fastly/fixtures/wasm_package/get.yaml
guyrichardson: Have removed from the metadata

guyrichardson:
> I think the binary file will be okay because that will not change, but the `update.yaml` will change frequently just by the nature of using `go-vcr`. I think this could easily inflate the git history, which over time will significantly slow down git operations. 

I've added recordIgnoreBody which adds a filter to set the body to "" in the recording. This seems to do the trick for the fixtures.

guyrichardson:
> I think the binary file will be okay because that will not change, but the `update.yaml` will change frequently just by the nature of using `go-vcr`. I think this could easily inflate the git history, which over time will significantly slow down git operations. This is a major concern because this is a library that a lot of users use and will frequently `go get`.

I've added recordIgnoreBody which adds a filter to the recorder which sets the body to "". This reduces down the fixture size to a normal size whilst still returning an appropriate response.

fastly/wasm.go
```diff
@@ -0,0 +1,102 @@
+package fastly
+
+import (
+	"fmt"
+	"io"
+	"time"
+)
+
+// WasmPackage is a container for data returned about a Wasm package.
+type WasmPackage struct {
```
phamann, line 10: On second review and thought, I think the public interface for this shouldn't mention Wasm at all. So this file should be renamed `package.go` and this struct `Package{}` and the methods `GetPackage()` etc. Apologies for the late  change, however this would better reflect the api endpoints and the terminology we use in C@E documentation etc.

PR #204, opened 2020-06-23
Set maximum dictionary/ACL sizes to 1,000

bendrucker:
This reduces the exported constants specifying the maximum sizes of ACLs and service dictionaries from 10,000 to 1,000, to match the documentation: 

> ACL containers are limited to 1000 ACL entries.

https://docs.fastly.com/en/guides/about-acls

> Dictionary containers are limited to 1000 items.

https://docs.fastly.com/en/guides/about-edge-dictionaries

This was originally introduced in https://github.com/fastly/go-fastly/pull/129 for use in Terraform, which is where I discovered this discrepancy. The behavior I observed was that changes appeared to succeed but the resulting ACL truncated excess entries.

### DIFF
--- a/fastly/fastly.go
+++ b/fastly/fastly.go
@@ -18,10 +18,10 @@ const (
 	BatchModifyMaximumOperations = 1000
 
 	// Represents the maximum number of items that can be placed within an Edge Dictionary.
-	MaximumDictionarySize = 10000
+	MaximumDictionarySize = 1000
 
 	// Represents the maximum number of entries that can be placed within an ACL.
-	MaximumACLSize = 10000
+	MaximumACLSize = 1000
 )
 
 type statusResp struct {


phamann:
Hey @bendrucker, thank you for the contribution. I'm afraid that the constants need to remain set to the upper hard limit of ACLs which is 10,000 within the go-fastly client. 

The default ACL limit is set to 1000 for customers however you can request an increase via Fastly support - of which the hard upper limit is 10000. Therefore, we use this constant in tests (for instance in the Terraform provider) to ensure logic works and we may have customers using the go-fastly client which have an increased limit that rely on the higher number.

However, It would be great if you can raise your issue against the Terraform provider repo, as that sounds like a bug in that logic, as we should probably first check your limits before attempting an apply with anything larger than your limit and shouldn't give a false positive of a successful apply with truncation.

bendrucker:
Sounds good, opened this more for discussion than to make this specific change, particularly b/c the original implementation referenced Terraform. Figured there might be soft/hard limits.

I'll write a failing acceptance test for the Terraform provider and go from there—thanks!

PR #206, opened 2020-07-01
Update CHANGELOG.md for v1.16.1

mccurdyc:
TL;DR
Updates changelog and bumps version for v1.16.1 release, which includes #184. #184
addresses existing linting issues, removes the `vendor` directory and improves
the developer workflow.

* [Rendered Changelog](https://github.com/fastly/go-fastly/blob/7cda1b8b04264b09ddcd688c64cc308601e9004e/CHANGELOG.md)
* [Rendered Diff](https://github.com/fastly/go-fastly/pull/206/files?short_path=4ac32a7#diff-4ac32a78649ca5bdd8e0ba38b7006a1e)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # Changelog
 
+## [v1.16.1](https://github.com/fastly/go-fastly/releases/tag/v1.16.1) (2020-07-07)
+
+[Full Changelog](https://github.com/fastly/go-fastly/compare/v1.16.0...v1.16.1)
+
+**Bug fixes:**
+
+- ci: add GH Action for fmt, vet, staticcheck, test [\#184](https://github.com/fastly/go-fastly/pull/184)
+
 ## [v1.16.0](https://github.com/fastly/go-fastly/releases/tag/v1.16.0) (2020-06-25)
 
 [Full Changelog](https://github.com/fastly/go-fastly/compare/v1.15.0...v1.16.0)


CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## [v0.16.0](https://github.com/fastly/go-fastly/releases/tag/v0.16.0) (2020-06-25)
```
mccurdyc, line 3: I'm not sure why it did this. The command that I ran was:

```bash
bash -c 'CHANGELOG_GITHUB_TOKEN=$(cat ~/.github/credentials) SEMVER_TAG="v1.16.1" make changelog'
```

Even when I re-run it, it says no diff. I'm assuming this is an issue with the erroneous release that I recently created.

mccurdyc: Even when I delete `v0.16.0` from my remote, the changelog fails to generate anything.

I don't expect the changelog to have enhancements or anything, just link to Full Changelog (i.e., the list of commits).

mccurdyc: @phamann Thoughts on deleting the erroneous `v0.16.0` tag and associated release?

phamann: Yep, think we should delete the 0.16.0 tag and release 👍

CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## [v1.16.1](https://github.com/fastly/go-fastly/releases/tag/v1.16.1) (2020-07-06)
+
+[Full Changelog](https://github.com/fastly/go-fastly/compare/v1.16.0...v1.16.1)
+
+**Bug fixes:**
```
phamann, line 7: Is this meant to be a Bug fix? More like Notes?

mccurdyc: I don't see a way to generate a Notes section using the github-changelog-generator and when I remove all tags from both of these, it doesn't generate a section for the latest release in the changelog.

Originally, I added 'bug fix' to the origin CI PR because it addressed the staticcheck issues (not exactly bugs).

PR #207, opened 2020-07-01
ci: install required tools for GitHub Action checks

mccurdyc:
Install required tooling in CI for checks.


mccurdyc:
@phamann Thoughts on removing Travis now that we have smoothed out the GitHub Action?

PR #210, opened 2020-07-09
Added support for /datacenters API call

joshuamsager:
Adds support for the following [utility API](https://developer.fastly.com/reference/api/utils/datacenter/).

The API `/datacenters` lists Fastly datacenters and their locations.

### DIFF
--- a/fastly/datacenters.go
+++ b/fastly/datacenters.go
@@ -1,12 +1,20 @@
 package fastly
 
+// Coordinates represent the location of a datacenter.
+type Coordinates struct {
+	Latitude   float64 `mapstructure:"latitude"`
+	Longtitude float64 `mapstructure:"longitude"`
+	X          float64 `mapstructure:"x"`
+	Y          float64 `mapstructure:"y"`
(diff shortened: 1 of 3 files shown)


fastly/datacenters.go
```diff
@@ -0,0 +1,24 @@
+package fastly
+
+// Datacenter is a list of Datacenters returned by the Fastly API.
+type Datacenter struct {
+	Code        string            `mapstructure:"code"`
+	Coordinates map[string]string `mapstructure:"coordinates"`
```
phamann, line 6: What is the reasoning for this being a `map[string][string]` and not a new type `Coordinates` struct?

joshuamsager: Simplicity, I can add a new `Coordinates` struct that matches the datatypes returned by the API

phamann:
@joshuamsager Apologies for the delay, this has now been released as [v1.17.0](https://github.com/fastly/go-fastly/releases/tag/v1.17.0)

PR #223, opened 2020-10-08
Adding WAF Rule Exclusion endpoints to go-fastly client

matpimenta:


fastly/waf_rule_exclusion.go
```diff
@@ -0,0 +1,297 @@
...
+	// The Web Application Firewall's version number.
+	WAFVersionNumber int
+	// Limit results to exclusions with the specified exclusions type.
+	FilterExclusionType *string
+	// Limit results to exclusions with the specified exclusion name.
+	FilterName *string
+	// Limit results to exclusions that represent the specified ModSecurity modsec_rule_id.
+	FilterModSedID *string
+	// Include relationships. Optional, comma-separated values. Permitted values: waf_rule_revision and waf_firewall_version.
+	Include *string
```
phamann, line 76: Same as above, this should be a `[]string`.

phamann (changes requested):
Looking good. One comment/change around an input API change from `string` to `[]sting`.

fastly/waf_rule_exclusion.go
```diff
@@ -0,0 +1,297 @@
...
+	// Limit results to exclusions with the specified exclusion name.
+	FilterName *string
+	// Limit results to exclusions that represent the specified ModSecurity modsec_rule_id.
+	FilterModSedID *string
+	// Limit the number of returned pages.
+	PageSize *int
+	// Request a specific page of exclusions.
+	PageNumber *int
+	// Include relationships. Optional, comma-separated values. Permitted values: waf_rule_revision and waf_firewall_version.
+	Include *string
```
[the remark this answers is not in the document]
matpimenta: Yes, it definitely looks nicer with a slice. I've updated the PR with this change

fastly/waf_rule_exclusion.go
matpimenta: Updated to slice as well

phamann (approved):
👍 LGTM

PR #225, opened 2020-10-20
Custom TLS API 

joel-g:
My organization needs Custom TLS certs and we hope to not maintain our own fork.

I may have discovered a platform bug. I have noticed that `POST /tls/certificates` will return a 500 Internal Server Error when the "type" field is blank (see https://developer.fastly.com/reference/api/tls/custom-certs/certificates/)

The Private Key (`POST /tls/private_keys`) and Platform TLS (`POST /tls/bulk/certificates`) APIs do not return an error if "type" is blank. 

This is relevant because the jsonapi marshaller only marshals the attributes and not the type. Therefore the CreateCustomCertificate() function in this PR does not currently work. I can work around this by adding "type" to the Go type and marshaling however this will break the convention found throughout the library.


joel-g:
I have also reported this possible bug in an enterprise support ticket: #300454

ryansouza:
Hey @joel-g. I believe I've found a fix for getting the correct `type` in the request json. In my testing adding a `primary` ID field to `CreateCustomCertificateInput` did cause the type to be set and did not require the ID to be set nor did it include an extraneous empty ID field in the json.

```
type CreateCustomCertificateInput struct {
	ID       string `jsonapi:"primary,tls_certificate"`
	CertBlob string `jsonapi:"attr,cert_blob"`
	Name     string `jsonapi:"attr,name"`
}
```

and from testing

```
// json before
{"data":{"type":"","attributes":{"cert_blob":"-----BEGIN CERTIFICATE-----\n...\n-----END CERTIFICATE-----\n","name":"My certificate"}}}

// json after
{"data":{"type":"tls_certificate","attributes":{"cert_blob":"-----BEGIN CERTIFICATE-----\n...\n-----END CERTIFICATE-----\n","name":"My certificate"}}}
```

Does this look like it will fix the issue?

joel-g:
@ryansouza I actually discovered the same thing this morning and added a "Type     string `jsonapi:"primary,tls_certificate"`" field that achieves the same thing.

I also have pushed more files to cover the /tls/activation and /tls/configuration API.

fastly/custom_tls_activation.go
```diff
@@ -0,0 +1,195 @@
...
+
+	return &a, nil
+}
+
+// CreateTLSActivationInput is used as input to the CreateTLSActivation function.
+type CreateTLSActivationInput struct {
+	TLSCertificate   *TLSCertificate   `jsonapi:"relation,tls_certificate,tls_certificate"`
+	TLSConfiguration *TLSConfiguration `jsonapi:"relation,tls_configuration,tls_configuration"`
+	TLSDomain        *TLSDomain        `jsonapi:"relation,tls_domain,tls_domain"`
+	Type             string            `jsonapi:"primary,tls_activation"` // Type value does not need to be set but existence of this key prevents server error due to API bug that requires "type" to be present.
```
thommahoney, line 121: The way [the code](https://github.com/google/jsonapi/blob/9246c912f5ff63f2e315efd188d254f93e5598f0/response.go#L286) is written, it can be named anything as long as it has `primary`. However, using `ID` is preferable here since it has [`omitempty`](https://github.com/google/jsonapi/blob/9246c912f5ff63f2e315efd188d254f93e5598f0/node.go#L39) and the value in this field will be marshaled there.

Also a quick note that the API requiring a `type` field is adherence to the [JSON-API.org specification](https://jsonapi.org/format/#crud-creating), for better or for worse:

> A resource can be created by sending a POST request to a URL that represents a collection of resources. The request MUST include a single resource object as primary data. The resource object MUST contain at least a type member.

ryansouza: Agree on the use of `ID` for consistency. I believe in the case of the Update struct these annotations can just be placed on the existing ID field and it will serialize as expected instead of having a `Type` field too

```
// UpdateTLSActivationInput is used as input to the UpdateTLSActivation function.
type UpdateTLSActivationInput struct {
	ID             string          `jsonapi:"primary,tls_activation"`
	TLSCertificate *TLSCertificate `jsonapi:"relation,tls_certificate,tls_certificate"`
}
```

joel-g: Thank you! I'll update per this feedback

joel-g: @thommahoney Noted re: JSON-API spec. I considered it a bug because /tls/configuration and other APIs do not return a server error if it is blank.

thommahoney: Understood, and thanks for clarifying here and in the code. 👍

joel-g:
Oh god what I have I done

fastly/custom_tls_activation_test.go
```diff
@@ -0,0 +1,165 @@
...
+		err = c.DeleteTLSActivation(&DeleteTLSActivationInput{
+			ID: ta.ID,
+		})
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func TestClient_CreateTLSActivation_validation(t *testing.T) {
```
phamann, line 96: This is not actually testing the validation business logic, only the happy path which we already have covered. It's more common for these functions to assert the error handling business logic of the methods, such as asserting that we returned a specific error message if an empty field was passed etc. Here is an example of what I mean: https://github.com/fastly/go-fastly/blob/master/fastly/datadog_test.go#L166-L182 

This applies generally to the other of validation tests elsewhere in this PR.

fastly/custom_tls_certificate.go
```diff
@@ -0,0 +1,195 @@
...
+import (
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/google/jsonapi"
+)
+
+// CustomCertificate represents a custom certificate. Uses common TLSDomain type from BulkCertificate
+type CustomCertificate struct {
```
phamann, line 12: To follow convention with other structures in this PR, this should be renamed to `CustomTLSCertificate` and the corresponding CRUD methods should also be renamed to match.
```suggestion
// CustomTLSCertificate represents a custom certificate. Uses common TLSDomain type from BulkCertificate.
type CustomTLSCertificate struct {
```

fastly/custom_tls_certificate.go
```diff
@@ -0,0 +1,195 @@
...
+	ID                 string       `jsonapi:"primary,tls_certificate"`
+	CreatedAt          *time.Time   `jsonapi:"attr,created_at,iso8601"`
```
phamann, line 14: To follow convention in the project do you mind moving this metadata field to the end of the struct.

fastly/custom_tls_certificate.go
```diff
@@ -0,0 +1,195 @@
...
+	IssuedTo           string       `jsonapi:"attr,issued_to"`
+	Issuer             string       `jsonapi:"attr,issuer"`
+	Name               string       `jsonapi:"attr,name"`
+	NotAfter           *time.Time   `jsonapi:"attr,not_after,iso8601"`
+	NotBefore          *time.Time   `jsonapi:"attr,not_before,iso8601"`
+	Replace            bool         `jsonapi:"attr,replace"`
+	SerialNumber       string       `jsonapi:"attr,serial_number"`
+	SignatureAlgorithm string       `jsonapi:"attr,signature_algorithm"`
+	UpdatedAt          *time.Time   `jsonapi:"attr,updated_at,iso8601"`
+	TLSDomains         []*TLSDomain `jsonapi:"relation,tls_domains,tls_domain"`
```
phamann, line 24: Duplication of type again, in case we miss any more can you check the other fields across the PR.

```suggestion
	TLSDomains         []*TLSDomain `jsonapi:"relation,tls_domains"`
```

fastly/custom_tls_certificate.go
```diff
@@ -0,0 +1,195 @@
...
+}
+
+// ListCustomCertificatesInput is used as input to the ListCustomCertificatesInput function.
+type ListCustomCertificatesInput struct {
+	PageNumber              *uint   // The page index for pagination.
+	PageSize                *uint   // The number of keys per page.
+	FilterTLSDomainsIDMatch *string // Filter certificates by their matching, fully-qualified domain name. Returns all partial matches. Must provide a value longer than 3 characters.
+	Sort                    *string // The order in which to list certificates. Valid values are created_at, not_before, not_after. May precede any value with a - for descending.
+}
```
phamann, line 33: This is missing some of the params, `FilterNotAfter` and `include`.

fastly/custom_tls_certificate.go
```diff
@@ -0,0 +1,195 @@
...
+	data, err := jsonapi.UnmarshalManyPayload(r.Body, reflect.TypeOf(new(CustomCertificate)))
+	if err != nil {
+		return nil, err
+	}
+
+	cc := make([]*CustomCertificate, len(data))
+	for i := range data {
+		typed, ok := data[i].(*CustomCertificate)
+		if !ok {
+			return nil, fmt.Errorf("got back a non-CustomCertificate response")
```
phamann, line 77: As per previous comment/convention.

```suggestion
			return nil, fmt.Errorf("unexpected response type: %T", data[i])
```

fastly/custom_tls_certificate.go
```diff
@@ -0,0 +1,195 @@
...
+	}
+
+	return &cc, nil
+}
+
+// CreateCustomCertificateInput is used as inpput to the CreateCustomCertificate function.
+type CreateCustomCertificateInput struct {
+	CertBlob string `jsonapi:"attr,cert_blob"`
+	Name     string `jsonapi:"attr,name"`
+	ID       string `jsonapi:"primary,tls_certificate"` // ID value does not need to be set.
```
phamann, line 115: Please move this ID to the top of the struct definition.

fastly/custom_tls_certificate.go
```diff
@@ -0,0 +1,195 @@
...
+
+	return &cc, nil
+}
+
+// UpdateCustomCertificateInput is used as inpput to the UpdateCustomCertificate function.
+type UpdateCustomCertificateInput struct {
+	ID       string `jsonapi:"primary,tls_certificate"`
+	CertBlob string `jsonapi:"attr,cert_blob"`
+	Name     string `jsonapi:"attr,name"`
+	Type     string `jsonapi:"primary,tls_certificate"`
```
phamann, line 148: Probably forgot to remove this after you added `ID` field with correct JSONAPI field tags.
```suggestion
```

fastly/custom_tls_certificate_test.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+func TestClient_CreateCustomCertificate_validation(t *testing.T) {
+	t.Parallel()
+
+	var err error
+	record(t, "custom_tls/create", func(c *Client) {
+		_, err = c.CreateCustomCertificate(&CreateCustomCertificateInput{
+			CertBlob: "-----BEGIN CERTIFICATE-----\n...\n-----END CERTIFICATE-----\n",
```
phamann, line 91: Out of interest, did you record this test with a valid cert and then redact afterwards? As I'm surprised the API accepted and returned a valid 200 for this request. (cc @ryansouza)

fastly/custom_tls_configuration.go
```diff
@@ -0,0 +1,151 @@
...
+import (
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/google/jsonapi"
+)
+
+// CustomTLSConfiguration represents a TLS configuration response from the Fastly API.
+type CustomTLSConfiguration struct {
```
phamann, line 12: This is missing the `tls_protocols` field, which is a slice of string.

fastly/custom_tls_configuration.go
```diff
@@ -0,0 +1,151 @@
...
+	ID            string       `jsonapi:"primary,tls_configuration"`
+	DNSRecords    []*DNSRecord `jsonapi:"relation,dns_records,dns_record"`
+	CreatedAt     *time.Time   `jsonapi:"attr,created_at,iso8601"`
```
phamann, line 15: Please move metadata fields to the bottom of the struct definition.

fastly/custom_tls_configuration.go
```diff
@@ -0,0 +1,151 @@
...
+
+// DNSRecord is a child of CustomTLSConfiguration
+type DNSRecord struct {
+	ID         string `jsonapi:"primary,dns_record"`
+	RecordType string `jsonapi:"attr,record_type"`
+	Region     string `jsonapi:"attr,region"`
+}
+
+// ListCustomTLSConfigurationsInput is used as input to the ListCustomTLSConfigurationsInput function.
+type ListCustomTLSConfigurationsInput struct {
```
phamann, line 31: This is missing the `include` field.

fastly/custom_tls_configuration.go
```diff
@@ -0,0 +1,151 @@
...
+	data, err := jsonapi.UnmarshalManyPayload(r.Body, reflect.TypeOf(new(CustomTLSConfiguration)))
+	if err != nil {
+		return nil, err
+	}
+
+	con := make([]*CustomTLSConfiguration, len(data))
+	for i := range data {
+		typed, ok := data[i].(*CustomTLSConfiguration)
+		if !ok {
+			return nil, fmt.Errorf("got back a non-tls-configuration response")
```
phamann, line 78:
```suggestion
			return nil, fmt.Errorf("unexpected response type: %T", data[i])
```

phamann (changes requested):
Hey @joel-g thank you for your contribution, this is a very large API with multiple endpoints, so we really appreciate your effort here. While there are a lot of comments, the majority are minor and just to enforce project convention and the others around how the JSONAPI library behaves. Let us know if you have any questions/issues.

(co-authored with @Integralist)

fastly/custom_tls_certificate_test.go
joel-g: You're right the API would not return 200 for this. I redacted/mocked following the convention I saw used for Platform TLS.

fastly/custom_tls_activation.go
```diff
@@ -0,0 +1,194 @@
...
+
+	return &a, nil
+}
+
+// CreateTLSActivationInput is used as input to the CreateTLSActivation function.
+type CreateTLSActivationInput struct {
+	ID               string            `jsonapi:"primary,tls_activation"` // ID value does not need to be set.
+	TLSCertificate   *TLSCertificate   `jsonapi:"relation,tls_certificate,tls_certificate"`
+	TLSConfiguration *TLSConfiguration `jsonapi:"relation,tls_configuration,tls_configuration"`
+	TLSDomain        *TLSDomain        `jsonapi:"relation,tls_domain,tls_domain"`
```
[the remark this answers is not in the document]
joel-g: I'm still new to jsonapi spec. I think at one point I thought it was required to have "relation,tls_domains,tls_domain" but later removed it everywhere but failed to remove it from this line.

fastly/custom_tls_activation.go
```diff
@@ -0,0 +1,194 @@
...
+
+	"github.com/google/jsonapi"
+)
+
+// TLSActivation represents a /tls/activations response.
+type TLSActivation struct {
+	ID               string            `jsonapi:"primary,tls_activation"`
+	TLSConfiguration *TLSConfiguration `jsonapi:"relation,tls_configuration"` // TLSConfiguration type shared with BulkCertificate
+	TLSDomain        *TLSDomain        `jsonapi:"relation,tls_domain"`        // TLSDomain type shared with BulkCertificate
+	TLSCertificate   *TLSCertificate   `jsonapi:"relation,tls_certificate"`
```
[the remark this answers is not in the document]
joel-g: Are you referring to the struct  field name or the types?

I named the types according to how they are named in the Postman collection: 

image

I named the struct field names following the json key names

I'm fine with either name. I was just following the most "official" names I could find.

fastly/custom_tls_configuration.go
joel-g: Good catch, added

fastly/custom_tls_configuration.go
joel-g: moved.

fastly/custom_tls_activation.go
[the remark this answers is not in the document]
phamann: The field names not the types, I agree that consistency with our API and docs are important however I like to focus on UX for the external internal fields.

fastly/custom_tls_configuration.go
joel-g: Added

fastly/custom_tls_activation.go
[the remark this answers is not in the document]
joel-g: Understood. It does feel redundant. I'll update them.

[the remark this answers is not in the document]
joel-g: Updated per feedback

fastly/custom_tls_activation.go
```diff
@@ -0,0 +1,194 @@
...
+		return nil, err
+	}
+
+	return &a, nil
+}
+
+// UpdateTLSActivationInput is used as input to the UpdateTLSActivation function.
+type UpdateTLSActivationInput struct {
+	ID             string          `jsonapi:"primary,tls_activation"`
+	TLSCertificate *TLSCertificate `jsonapi:"relation,tls_certificate,tls_certificate"`
```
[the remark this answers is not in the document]
joel-g: @phamann My understanding from the Postman collection was that only the Certificate would be updated.
image

fastly/custom_tls_certificate.go
joel-g: Done!

fastly/custom_tls_certificate.go
joel-g: Updated

fastly/custom_tls_activation.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+// UpdateTLSActivation
+func (c *Client) UpdateTLSActivation(i *UpdateTLSActivationInput) (*TLSActivation, error) {
+	if i.ID == "" {
+		return nil, ErrMissingID
+	}
+	if i.TLSCertificate == nil {
+		return nil, ErrMissingTLSCertificate
+	}
```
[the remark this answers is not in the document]
joel-g: Added

fastly/custom_tls_activation_test.go
```diff
@@ -0,0 +1,165 @@
...
+		gta, err = c.GetTLSActivation(&GetTLSActivationInput{
+			ID: ta.ID,
+		})
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if ta.ID != gta.ID {
+		t.Errorf("bad ID: %q (%q)", ta.ID, gta.ID)
+	}
```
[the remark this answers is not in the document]
joel-g: Added

fastly/custom_tls_activation_test.go
joel-g: That makes sense. I will add these. I was following platform_tls_test.go and it does not have these "not happy" paths tested.

fastly/custom_tls_activation.go
```diff
@@ -0,0 +1,194 @@
...
+
+	return &a, nil
+}
+
+// CreateTLSActivationInput is used as input to the CreateTLSActivation function.
+type CreateTLSActivationInput struct {
+	ID               string            `jsonapi:"primary,tls_activation"` // ID value does not need to be set.
+	TLSCertificate   *TLSCertificate   `jsonapi:"relation,tls_certificate,tls_certificate"`
+	TLSConfiguration *TLSConfiguration `jsonapi:"relation,tls_configuration,tls_configuration"`
```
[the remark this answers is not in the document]
joel-g: Done

fastly/custom_tls_certificate.go
joel-g: Done

fastly/custom_tls_certificate.go
joel-g: Done

fastly/custom_tls_certificate.go
joel-g: Done

fastly/custom_tls_certificate.go
joel-g: Fixed.

joel-g:
@phamann thank you for the thorough review. I believe I have resolved all of the requested changes except this one: https://github.com/fastly/go-fastly/pull/225#discussion_r518736212

Can you confirm that the official Postman collection is incomplete and that I should add Configuration and Domain fields?

fastly/custom_tls_activation.go
```diff
@@ -0,0 +1,198 @@
...
+// TLSActivation represents a /tls/activations response.
+type TLSActivation struct {
+	ID            string                `jsonapi:"primary,tls_activation"`
+	Configuration *TLSConfiguration     `jsonapi:"relation,tls_configuration"` // TLSConfiguration type shared with BulkCertificate
+	Domain        *TLSDomain            `jsonapi:"relation,tls_domain"`        // TLSDomain type shared with BulkCertificate
+	Certificate   *CustomTLSCertificate `jsonapi:"relation,tls_certificate"`
+	CreatedAt     *time.Time            `jsonapi:"attr,created_at,iso8601"`
+}
+
+// TLSCertificate represents a certificate relationship. See CustomTLSCertificate for the /tls/certificates API.
```
phamann, line 20: Missed this.

```suggestion
```

fastly/custom_tls_activation.go
```diff
@@ -0,0 +1,198 @@
...
+		return nil, err
+	}
+
+	return &a, nil
+}
+
+// UpdateTLSActivationInput is used as input to the UpdateTLSActivation function.
+type UpdateTLSActivationInput struct {
+	ID          string                `jsonapi:"primary,tls_activation"`
+	Certificate *CustomTLSCertificate `jsonapi:"relation,tls_certificate,tls_certificate"` // Only ID of CustomTLSCertificate needs to be set.
```
phamann, line 159: Repeat of `tls_certificate`.

```suggestion
	Certificate *CustomTLSCertificate `jsonapi:"relation,tls_certificate"` // Only ID of CustomTLSCertificate needs to be set.
```

fastly/custom_tls_activation.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// UpdateTLSActivation
```
phamann, line 162:
```suggestion
// UpdateTLSActivation updates the certificate used to terminate TLS traffic for the domain associated with this TLS activation.
```

fastly/custom_tls_certificate.go
```diff
@@ -0,0 +1,196 @@
...
+
+import (
+	"fmt"
+	"reflect"
+	"time"
+
+	"github.com/google/jsonapi"
+)
+
+// CustomTLSCertificate represents a custom certificate. Uses common TLSDomain type from BulkCertificate
```
phamann, line 11:
```suggestion
// CustomTLSCertificate represents a custom certificate. Uses common TLSDomain type from BulkCertificate.
```

fastly/custom_tls_certificate.go
phamann: This is still here :)

phamann (changes requested):
@joel-g Thanks for the updates, this is looking good, only a couple of minor things missed and this should be good to merge.

fastly/custom_tls_activation.go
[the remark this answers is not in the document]
phamann: 👍 Yes, you're correct. The API docs also seem to agree with the postman collection: https://developer.fastly.com/reference/api/tls/custom-certs/activations/#update-tls-activation

phamann (approved):
LGTM :tada:

Thank you SO MUCH for your contribution @joel-g and for your patience and understanding throughout this process. Apologies it's taken so long. Part of the reason has been we're transitioning the go-fastly project over to a new internal team and owner, @Integralist, which will hopefully mean there won't be as much delay in the future.

Related: I want him to have one final look over the PR and then we will merge and release this week for you.

joel-g:
@phamann that was actually very fast compared so some vendors open source. And it was a big PR so I'm happy to "get it right". I'm already working on adding these to https://github.com/fastly/terraform-provider-fastly so I'll see you there

fastly/custom_tls_certificate.go
```diff
@@ -0,0 +1,195 @@
...
+	IssuedTo           string       `jsonapi:"attr,issued_to"`
+	Issuer             string       `jsonapi:"attr,issuer"`
+	Name               string       `jsonapi:"attr,name"`
+	NotAfter           *time.Time   `jsonapi:"attr,not_after,iso8601"`
+	NotBefore          *time.Time   `jsonapi:"attr,not_before,iso8601"`
+	Replace            bool         `jsonapi:"attr,replace"`
+	SerialNumber       string       `jsonapi:"attr,serial_number"`
+	SignatureAlgorithm string       `jsonapi:"attr,signature_algorithm"`
+	TLSDomains         []*TLSDomain `jsonapi:"relation,tls_domains"`
+	CreatedAt          *time.Time   `jsonapi:"attr,created_at,iso8601"`
```
Integralist, line 23: Was going to ask for the jsonapi 'attributes' to be grouped and ordered, but I then recalled @phamann telling me this is a pre-existing convention used, where 'metadata' related fields are grouped at the bottom of the struct. 

Personally I'm not sure if I'm a fan of it, as I find it disorientating when comparing the code to the actual API documentation, but for the sake of forward progress I'm not going to 'bikeshed' this change 🙂

fastly/custom_tls_configuration.go
```diff
@@ -0,0 +1,153 @@
...
+			result[key] = fmt.Sprintf("%v", reflect.ValueOf(value).Elem())
+		}
+	}
+	return result
+}
+
+// ListCustomTLSConfigurations list all TLS configurations.
+func (c *Client) ListCustomTLSConfigurations(i *ListCustomTLSConfigurationsInput) ([]*CustomTLSConfiguration, error) {
+	p := "/tls/configurations"
+	filters := &RequestOptions{
```
Integralist, line 60: to be consistent with other files can we change from `filters` to `ro` (will require changes to references elsewhere in this file) also fits more closely with Go's idiom of naming variables after the associated type.

```suggestion
	ro := &RequestOptions{
```

fastly/custom_tls_configuration.go
```diff
@@ -0,0 +1,153 @@
...
+		Params: i.formatFilters(),
+		Headers: map[string]string{
+			"Accept": "application/vnd.api+json", // this is required otherwise the filters don't work
+		},
+	}
+
+	r, err := c.Get(p, filters)
```
Integralist, line 67: based on my earlier comment...

```suggestion
	r, err := c.Get(p, ro)
```

fastly/custom_tls_configuration.go
```diff
@@ -0,0 +1,153 @@
...
+		con[i] = typed
+	}
+
+	return con, nil
+}
+
+// GetCustomTLSConfigurationInput is used as input to the GetCustomTLSConfiguration function.
+type GetCustomTLSConfigurationInput struct {
+	ID      string
+	Include *string // Include related objects. Optional, comma-separated values. Permitted values: dns_records.
```
Integralist, line 92: I might just be confused 🙂  but could this probably just be a non-pointer string? due to the fact this isn't a write/update operation. 

e.g. we're typically using pointers for basic types that are optional fields because of the problem inherent to a marshal of the data when the field is purposely given a value that aligns with the field type's zero value. 

But that's not relevant in this case I think because it's just a READ operation.

fastly/custom_tls_configuration.go
```diff
@@ -0,0 +1,153 @@
...
+	if err := jsonapi.UnmarshalPayload(r.Body, &con); err != nil {
+		return nil, err
+	}
+
+	return &con, nil
+}
+
+// UpdateCustomTLSConfigurationInput is used as input to the UpdateCustomTLSConfiguration function.
+type UpdateCustomTLSConfigurationInput struct {
+	ID   string `jsonapi:"attr,id"`
```
Integralist, line 128: Does `ID` require the jsonapi struct tag or can it be omitted? as the API docs for this endpoint suggests only `name` is a required attribute.

fastly/errors.go
```diff
@@ -149,6 +149,18 @@ var ErrMissingCertBlob = errors.New("missing required field 'CertBlob'")
...
 
+// ErrMissingTLSCertificate is an error that is returned from an input struct that requires
+// a "TLSCertificate" struct, but one was not set.
+var ErrMissingTLSCertificate = errors.New("missing required field 'TLSCertificate'")
+
+// ErrMissingTLSConfiguration is an error that is returned from an input struct that requires
+// a "TLSConfiguration" struct, but one was not set.
+var ErrMissingTLSConfiguration = errors.New("missing required field 'TLSConfiguration'")
+
+// ErrMissingTLSDomainID is an error that is returned from an input struct that requires
```
Integralist, line 160: Looks like this should have been updated when the variable name was updated...

```suggestion
// ErrMissingTLSDomain is an error that is returned from an input struct that requires
```

fastly/errors.go
```diff
@@ -149,6 +149,18 @@ var ErrMissingCertBlob = errors.New("missing required field 'CertBlob'")
...
+// a "TLSDomain" struct, but one was not set.
+var ErrMissingTLSDomain = errors.New("missing required field 'TLSDomainID'")
```
Integralist, line 162: Looks like an old reference...

```suggestion
var ErrMissingTLSDomain = errors.New("missing required field 'TLSDomain'")
```

Integralist (commented):
Thanks for the great work on this PR @joel-g 🙂  it's a huge chunk of the TLS API so we really appreciate the effort you've put into this. 

Before I approve the PR I just had some small comments for you. If you could take a look and either address/resolve them or let me know your thoughts.

fastly/custom_tls_certificate.go
joel-g: Yeah. I originally put them in the order of the JSON API (which I believe is alphabetical) but @phamann asked me to move the CreatedAt and UpdatedAt fields to the bottom.

fastly/custom_tls_activation.go
```diff
@@ -0,0 +1,196 @@
...
+		Headers: map[string]string{
+			"Accept": "application/vnd.api+json", // this is required otherwise the params don't work
+		},
+	}
+
+	if i.Include != nil {
+		ro.Params = map[string]string{"include": *i.Include}
+	}
+
+	r, err := c.Get(p, nil)
```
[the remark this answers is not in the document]
joel-g: 🤦  fixed

fastly/custom_tls_configuration.go
joel-g: Updated

fastly/custom_tls_configuration.go
joel-g: You're right. The ID is not marshaled into the body, it just used in the path. Removed.

fastly/custom_tls_activation.go
```diff
@@ -0,0 +1,196 @@
...
+	ID            string                `jsonapi:"primary,tls_activation"`
+	Configuration *TLSConfiguration     `jsonapi:"relation,tls_configuration"` // TLSConfiguration type shared with BulkCertificate
+	Domain        *TLSDomain            `jsonapi:"relation,tls_domain"`        // TLSDomain type shared with BulkCertificate
+	Certificate   *CustomTLSCertificate `jsonapi:"relation,tls_certificate"`
+	CreatedAt     *time.Time            `jsonapi:"attr,created_at,iso8601"`
+}
+
+// ListTLSActivationsInput is used as input to the ListTLSActivations function.
+type ListTLSActivationsInput struct {
+	FilterTLSCertificateID   *string // Limit the returned activations to a specific certificate.
```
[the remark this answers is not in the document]
joel-g: I was following the pattern I observed of having optional fields be pointers. I didn't catch that it was only on write operations. I can change this if that's what you decide.

fastly/custom_tls_configuration.go
joel-g: That makes sense. Updated.

fastly/custom_tls_activation.go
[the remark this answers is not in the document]
Integralist: @joel-g to give you some context, take a read of https://willnorris.com/2014/05/go-rest-apis-and-pointers/ -- from my understanding the issue of not using pointer references are only relevant to write/update operations, where as create/read operations are less of a concern for us (I could be mistaken of course, but that was my takeaway).

fastly/custom_tls_configuration.go
Integralist: Looks like the struct tag is still showing, maybe you need to re-push?

fastly/custom_tls_configuration.go
Integralist: I think you might need to re-push as I can still see a pointer reference used.

fastly/custom_tls_activation.go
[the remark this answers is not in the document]
Integralist: FYI - I'm PTO tomorrow. So I'll check back in on this PR Monday. Have a great weekend and thank you for your contributions so far @joel-g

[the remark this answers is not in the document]
joel-g: I read the blog post. If I make "FilterTLSCertificateID" a string rather than a *string will it not filter on an empty string and return nothing when the value is not set?

I see in the Bulk TLS API it is using pointers:

https://github.com/fastly/go-fastly/blob/57854aff2f13a1958bb496db0edd9a491c465a79/fastly/platform_tls.go#L39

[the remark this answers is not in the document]
Integralist: Hi @joel-g apologies for the long delay in responding. 

I've had a few other projects I needed to make progress on and we also had a sudden push to get a long standing v2.0.0 branch for Go-Fastly released:
https://github.com/fastly/go-fastly/releases/tag/v2.0.0

Please see the [new migration notes](https://github.com/fastly/go-fastly#migrating-from-v1-to-v2) in the README which details the main changes (there shouldn't be anything surprising or problematic in there, just working towards a more consistent code base). 

So if you could please rebase in the `master` branch when you get a chance please.

With regards to your question...

> If I make "FilterTLSCertificateID" a string rather than a *string will it not filter on an empty string and return nothing when the value is not set?

Correct.

But as this is a _list_ instruction (in other words a READ only operation) then there is no break in expected behaviour (i.e. a consumer of this code will have to choose not to provide that information). Now compare that to omitting a non-pointer field when carrying out a WRITE operation... in that case a missing value could result in a fundamentally different behaviour.

So we're very much in favour of pointer references for optional fields when doing WRITE operations (i.e. update structs) but less concerned when it comes to READ/CREATE operations (i.e. create structs).

Hope that makes sense.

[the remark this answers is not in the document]
joel-g: Ok thanks @Integralist, I have pulled in the changes from master and updated the three List operations introduced in this PR to no longer use pointers.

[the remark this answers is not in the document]
joel-g: @Integralist @phamann any updates on this?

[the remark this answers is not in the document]
Integralist: Hi @joel-g thanks for your patience. I've been out on paternity leave but I'm back now and am focused on reviewing your PR and helping it over the finish line.

Integralist (commented):
Hi @joel-g a few quick comments.

Integralist (changes requested):
A few more small tweaks required please Joel. Thanks!

fastly/custom_tls_certificate.go
```diff
@@ -0,0 +1,205 @@
...
+	UpdatedAt          *time.Time   `jsonapi:"attr,updated_at,iso8601"`
+}
+
+// ListCustomTLSCertificatesInput is used as input to the ListCustomTLSCertificatesInput function.
+type ListCustomTLSCertificatesInput struct {
+	FilterNotAfter          string // Limit the returned certificates to those that expire prior to the specified date in UTC. Accepts parameters: lte (e.g., filter[not_after][lte]=2020-05-05).
+	FilterTLSDomainsIDMatch string // Filter certificates by their matching, fully-qualified domain name. Returns all partial matches. Must provide a value longer than 3 characters.
```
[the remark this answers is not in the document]
joel-g: The comments are carried over from the descriptions I found in the Fastly Postman collection. The function ListCustomTLSCertificates() won't send an empty string due to the formatFilters() function skipping "" but I will add a test to validate that.

fastly/custom_tls_certificate.go
```diff
@@ -0,0 +1,205 @@
...
+	PageSize                int    // The number of keys per page.
+	Sort                    string // The order in which to list certificates. Valid values are created_at, not_before, not_after. May precede any value with a - for descending.
+}
+
+// formatFilters converts user input into query parameters for filtering.
+func (i *ListCustomTLSCertificatesInput) formatFilters() map[string]string {
+	result := map[string]string{}
+	pairings := map[string]interface{}{
+		"filter[not_after]":             i.FilterNotAfter,
+		"filter[tls_domains.id][match]": i.FilterTLSDomainsIDMatch,
```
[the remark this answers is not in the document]
joel-g: Fixed!

fastly/custom_tls_certificate.go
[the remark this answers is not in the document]
joel-g: Actually, I believe the case of blank filters is already tested.

fastly/errors.go
```diff
@@ -169,6 +169,18 @@ var ErrMissingCertBlob = errors.New("missing required field 'CertBlob'")
 // a "IntermediatesBlob" key, but one was not set.
 var ErrMissingIntermediatesBlob = errors.New("missing required field 'IntermediatesBlob'")
 
+// ErrMissingTLSCertificate is an error that is returned from an input struct that requires
+// a "TLSCertificate" struct, but one was not set.
+var ErrMissingTLSCertificate = errors.New("missing required field 'TLSCertificate'")
+
+// ErrMissingTLSConfiguration is an error that is returned from an input struct that requires
+// a "TLSConfiguration" struct, but one was not set.
```
Integralist, line 177: As I discuss above, a suggested change could be...

```suggestion
// a "Configuration" field assigned a "TLSConfiguration" struct, but one was not set.
```

fastly/errors.go
```diff
@@ -169,6 +169,18 @@ var ErrMissingCertBlob = errors.New("missing required field 'CertBlob'")
...
+var ErrMissingTLSConfiguration = errors.New("missing required field 'TLSConfiguration'")
+
+// ErrMissingTLSDomain is an error that is returned from an input struct that requires
+// a "TLSDomain" struct, but one was not set.
```
Integralist, line 181: As I discuss above, a suggested change could be...

```suggestion
// a "Domain" field assigned a "TLSDomain" struct, but one was not set.
```

Integralist (commented):
Apologies @joel-g I think some of my suggestions might have been missed/mis-understood. I've added/re-added some stuff.

Integralist (approved):
OK I think we might have gotten there 🎉

Thanks for all your patience and hard work @joel-g 👍🏻

> **Note**: I have three  other Go-Fastly PRs that I'm reviewing today, and so I might not cut a new Go-Fastly release just yet (e.g. once merging this PR) as I'd like to combine the other PR changesets into a single release.

PR #233, opened 2020-11-17
README corrections as per 2.0.0 release

Integralist:
**Problem**: the pkg.dev.go path wasn't what I had expected it to be + I neglected the imports in the example code.


phamann (approved):
👍🏻

PR #235, opened 2020-12-08
Add support for compression_codec to logging file sink endpoints

kellymclaughlin:
Add support for compression_codec to the following logging endpoints in go-fastly:

- Azure Blob
- FTP
- GCS
- OpenStack
- S3
- SFTP

`gzip_level` is still supported, but `compression_codec` is now the preferred option. The intention is that if `gzip_level` is present it overrides any `compression_codec` value.


jaredmorrow (commented):
This looks good to me, but I'll let someone from the owning team chime in.

Integralist:
Thanks for the PR! 

There's a couple of other PRs from @kellymclaughlin that I need to catch back up on (since I've been on paternity leave), so those will take priority for me today. Hopefully I can find some time later today to review this PR though.

Integralist:
@kellymclaughlin @jaredmorrow am I right in noticing that the new field `compression_codec` isn't documented at developer.fastly.com under our api reference. I presume there plans to ensure it's added there once this PR is merged.

fastly/blobstorage_test.go
```diff
@@ -15,21 +15,21 @@ func TestClient_BlobStorages(t *testing.T) {
...
-			ServiceID:       testServiceID,
-			ServiceVersion:  tv.Number,
-			Name:            "test-blobstorage",
-			Path:            "/logs",
-			AccountName:     "test",
-			Container:       "fastly",
-			SASToken:        "sv=2015-04-05&ss=b&srt=sco&sp=rw&se=2019-07-21T18%3A00%3A00Z&sig=3ABdLOJZosCp0o491T%2BqZGKIhafF1nlM3MzESDDD3Gg%3D",
-			Period:          12,
-			TimestampFormat: "%Y-%m-%dT%H:%M:%S.000",
-			GzipLevel:       9,
```
Integralist, line 27: re:

> The intention is that if gzip_level is present it overrides any compression_codec value.

I think we need a test that validates when `GzipLevel` _and_ `CompressionCodec` are provided, that `GzipLevel` will take priority.

Let's ensure we have that test scenario defined for all the updated test files.

That said, I don't think I saw any logic in this change-set for implementing GzipLevel taking priority over CompressionCodec, so could you clarify how that's achieved 🤔

kellymclaughlin: | That said, I don't think I saw any logic in this change-set for implementing GzipLevel taking priority over CompressionCodec, so could you clarify how that's achieved :thinking:

This prioritization is handled by the API code so we just exercise it here.

kellymclaughlin: Ok, I expanded the testing so each applicable logging endpoint test suite exercises the interaction between these two options.

Integralist (approved):

PR #238, opened 2020-12-14
Paginate services

kilgarth:
Paginate the `/service` endpoint

### DIFF
--- a/fastly/fixtures/services/list.yaml
+++ b/fastly/fixtures/services/list.yaml
@@ -10,7 +10,7 @@ interactions:
       - xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
       User-Agent:
       - FastlyGo/0.4.3.dev (+github.com/fastly/go-fastly; go1.12)
-    url: https://api.fastly.com/service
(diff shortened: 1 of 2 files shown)


Integralist (commented):
Hi @kilgarth thanks for opening this PR, unfortunately we won't be able to accept these changes for the following reasons:

1. Not all users require this behaviour, so modifying the existing `ListServices` method would break backwards compatibility and require bumping to a new major version release. 
2. We should also continue to follow existing patterns in the library and allow pagination via input values (not implicitly fetching all).
3. Even if we tried to move this behaviour to a new method (e.g. `ListServicesAll`) it could potentially exhaust a program's memory (as the amount of data being handled in-memory could be unbounded).

I'll leave this PR open for the moment so you have an opportunity to discuss alternatives. Otherwise if no further comments are left by the end of the week I'll go ahead and close this PR.

Thanks again for taking the time to implement new functionality to Go-Fastly. It's appreciated.

Integralist:
As I mentioned a few days ago: there's been no further discussion so I'm going to go ahead and close this PR.

PR #151, opened 2020-04-07
Active version details returned by service/:id/details

ovv:
Some information is not made accessible from the `service/:id/details` request. This is useful to reduce the number of overall API calls made to fastly.

This change is not perfectly backward compatible as `ServiceDetail.ActiveVersion` changes type. If this is not acceptable we can figure out an alternative.

PR #239, opened 2020-12-17
Create new error abstraction for field validation

Integralist:
Create new error abstraction for field validation. This helps with consistency and maintainability.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ The move to more consistent field names in some cases will have resulted in the 
 
 The change in type for [basic types](https://tour.golang.org/basics/11) that are optional on input structs related to write/update operations is designed to avoid unexpected behaviours when dealing with their zero value (see [this reference](https://willnorris.com/2014/05/go-rest-apis-and-pointers/) for more details). As part of this change we now provide [helper functions](./fastly/basictypes_helper.go) to assist with generating the new pointer types required.
 
+> Note: some read/list operations require fields to be provided but if omitted a zero value will be used when marshaling the data structure into JSON. This too can cause confusion, which is why some input structs define their mandatory fields as pointers (to ensure that the backend can distinguish between a zero value and an omitted field).
+
 ## Examples
 
 Fastly's API is designed to work in the following manner:
--- a/README.md
+++ b/README.md
@@ -1,9 +1,10 @@
 # Go Fastly
 
-[![Go Documentation](http://img.shields.io/badge/go-documentation-blue.svg?style=flat-square)][godocs]
+[![Go Documentation](http://img.shields.io/badge/go-documentation-blue.svg?style=flat-square)][latest]
 
-[godocs]: https://pkg.go.dev/github.com/fastly/go-fastly/fastly?tab=doc
-[v2]: https://pkg.go.dev/github.com/fastly/go-fastly/v2
+[latest]: https://pkg.go.dev/github.com/fastly/go-fastly/v3/fastly
+[v3]: https://pkg.go.dev/github.com/fastly/go-fastly/v3/fastly
+[v2]: https://pkg.go.dev/github.com/fastly/go-fastly/v2/fastly
 [v1]: https://pkg.go.dev/github.com/fastly/go-fastly
 
 Go Fastly is a Golang API client for interacting with most facets of the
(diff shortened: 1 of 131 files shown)


README.md
```diff
@@ -31,6 +31,17 @@ The move to more consistent field names in some cases will have resulted in the
 
 The change in type for [basic types](https://tour.golang.org/basics/11) that are optional on input structs related to write/update operations is designed to avoid unexpected behaviours when dealing with their zero value (see [this reference](https://willnorris.com/2014/05/go-rest-apis-and-pointers/) for more details). As part of this change we now provide [helper functions](./fastly/basictypes_helper.go) to assist with generating the new pointer types required.
 
+> Note: some read/list operations require fields to be provided but if omitted a zero value will be used when marshaling the data structure into JSON. This too can cause confusion, which is why some input structs define their mandatory fields as pointers (to ensure that the backend can distinguish between a zero value and an omitted field).
+
+## Migrating from v2 to v3
```
Integralist, line 36: @phamann I added this to the README because I realized that these changes were potentially a breaking change for consumers of Go-Fastly so it should probably be a v3.0.0 release 😬 

That said, I don't think we should necessarily be afraid of major version releases. I know it would be nice to have a major release contain something 'juicy' but the fact is if something should be fixed for sustainability purposes then we shouldn't feel like we have to wait to bump a major version.

As a side note if we go to v3 then we could look to incorporate this PR https://github.com/fastly/go-fastly/pull/151 which I believe would be a breaking change as well.

README.md
```diff
@@ -1,9 +1,10 @@
 # Go Fastly
```
Integralist, line 1: @phamann updated README as per our conversation.

go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/fastly/go-fastly/v2
+module github.com/fastly/go-fastly/v3
```
Integralist, line 1: @phamann I've also bumped to v3 in the go.mod

fastly/errors.go
```diff
@@ -9,180 +9,44 @@ import (
...
-var ErrMissingKMSKeyID = errors.New("missing required field 'ServerSideEncryptionKMSKeyID'")
-
-// ErrMissingCertBlob is an error that is returned from an input struct that requires
-// a "CertBlob" key, but one was not set.
-var ErrMissingCertBlob = errors.New("missing required field 'CertBlob'")
+// FieldError represents a custom error type for API data fields.
+type FieldError struct {
+	kind    string
+	message string
+}
```
phamann, line 16: While I agree that a new custom error type `FieldError` is a good and needed abstraction to the problem; I'm slightly confused as to why you've removed the package level constants to store the unique instances of the errors? 

We now have 100+ instances of `NewFieldError("ServiceID")` across the codebase instead of a single constant definition in this file which can then be referenced elsewhere (and reduce the memory allocation significantly). The bigger issue is we then are forced to assert against the raw error string in every test, meaning we now have:
```diff
-      if err != ErrMissingServiceVersion {
+      if err.Error() != "missing required field 'ServiceID'" {
```
 in many places which I find very brittle compared to the previous implementation. I.e. if we chose to change the implementation of `FieldError` to return a slightly different message at a later date all of these tests will instantly fail.

Therefore, I'd suggest we go with `FieldError` but keep the constants and their references which should dramatically reduce the change set of this PR. Alternatively if you want to improve the error assertions accross the codebase, you could use the fancy new [`error.Is()`](https://golang.org/pkg/errors/#Is) introduced in Go 1.13.

Unless I'm missing something and there was a good reason you when with this pattern of a new instance of `FieldError` per file?

README.md
phamann: Makes sense 👍🏻 
Also +1 to incorporating any other breaking change PRs if we are to bump the version.

phamann (changes requested):
Looks good, other than my comment around error constants, let me know if you want to chat about the best pattern before changing anything.

Integralist:
@phamann changes made, waiting re-review

phamann (approved):
👍🏻 LGTM

Will you be incorporating the other PR into this release?

Integralist:
@phamann yes. the author [replied to me eariler](https://github.com/fastly/go-fastly/pull/151#issuecomment-754715458) to say they would take a look next week.

So with that in mind. I can merge this change but not cut a release until after they've managed to wrap up any other changes and write some tests in their PR.

Integralist:
@fastly/developer-relations @phamann  I've reverted a bunch of approved changes (see the latest commits above). I'm going to open a few separate PRs to home these changes (as they've already been approved by Patrick they should be quick to review/approve) but it'll help ensure the dynamically generated CHANGELOG will have explicit callouts to each change (whereas if I kept them all within this PR they would be lost).

PR #241, opened 2021-01-05
Add basic managed logging endpoint support to go-fastly

jaredmorrow:
Support enabling and disabling managed logging for a service.

### DIFF
--- a/fastly/fixtures/managed_logging/delete.yaml
+++ b/fastly/fixtures/managed_logging/delete.yaml
@@ -19,11 +19,11 @@ interactions:
       Content-Type:
       - application/json
       Date:
-      - Thu, 07 Jan 2021 19:28:05 GMT
+      - Mon, 11 Jan 2021 14:05:09 GMT
       Fastly-Ratelimit-Remaining:
-      - "19995"
+      - "19996"
       Fastly-Ratelimit-Reset:
-      - "1610049600"
+      - "1610377200"
       Status:
       - 200 OK
       Strict-Transport-Security:
@@ -37,9 +37,9 @@ interactions:
       X-Cache-Hits:
       - 0, 0
       X-Served-By:
-      - cache-control-slwdc9037-CONTROL-SLWDC, cache-den8229-DEN
+      - cache-control-slwdc9036-CONTROL-SLWDC, cache-den8275-DEN
       X-Timer:
-      - S1610047685.823978,VS0,VE866
+      - S1610373908.316897,VS0,VE892
     status: 200 OK
     code: 200
     duration: ""
--- a/fastly/fixtures/managed_logging/delete.yaml
+++ b/fastly/fixtures/managed_logging/delete.yaml
@@ -1,0 +1,45 @@
+---
+version: 1
+interactions:
+- request:
+    body: ""
+    form: {}
+    headers:
+      User-Agent:
+      - FastlyGo/2.1.0 (+github.com/fastly/go-fastly; go1.15.5)
+    url: https://api.fastly.com/service/7i6HN3TK9wS159v2gPAZ8A/log_stream/managed/instance_output
+    method: DELETE
+  response:
+    body: '{"status":"ok"}'
+    headers:
+      Accept-Ranges:
+      - bytes
+      Cache-Control:
+      - no-cache
+      Content-Type:
+      - application/json
+      Date:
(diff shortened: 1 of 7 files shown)


Integralist:
@jaredmorrow I'm guessing this error is because you need to have that feature enabled on your account for the VCR fixture to work?

```
--- FAIL: TestClient_ManagedLogging (0.21s)
    managed_logging_test.go:27: 401 - Unauthorized:
        
            Title:  Provided credentials are missing or invalid
```

jaredmorrow:
@Integralist That was my initial concern / thought as well.  I didn't think it was needed, but I still added the test service `7i6HN3TK9wS159v2gPAZ8A` to the feature in question in Fastly App -> Admin.

jaredmorrow:
@Integralist it was a case issue.  I switched to underscore to match the rest of the fixtures.  Now it works.

jaredmorrow:
I'm going to add support for other types of managed logging, so I'm converting to a draft for now.

fastly/fixtures/managed_logging/recreate.yaml
```diff
@@ -0,0 +1,51 @@
...
+      - 7i6HN3TK9wS159v2gPAZ8A
+    headers:
+      Content-Type:
+      - application/x-www-form-urlencoded
+      User-Agent:
+      - FastlyGo/2.1.0 (+github.com/fastly/go-fastly; go1.15.5)
+    url: https://api.fastly.com/service/7i6HN3TK9wS159v2gPAZ8A/log_stream/managed/instance_output
+    method: POST
+  response:
+    body: '{"msg":"Event Buffer has already been enabled for this service"}'
```
Integralist, line 19: My understanding is that the production name is now 'Log Tailing':

```suggestion
    body: '{"msg":"Log Tailing has already been enabled for this service"}'
```

fastly/managed_logging.go
```diff
@@ -0,0 +1,97 @@
...
+		Kind ManagedLoggingKind
+	}
+)
+
+const (
+	ManagedLoggingUnset ManagedLoggingKind = iota
+	ManagedLoggingInstanceOutput
+)
+
+// CreateManagedLogging enabled managed logging for a service.
```
Integralist, line 32:
```suggestion
// CreateManagedLogging enables managed logging for a service.
```

fastly/fixtures/managed_logging/delete.yaml
```diff
@@ -0,0 +1,45 @@
+---
+version: 1
+interactions:
+- request:
+    body: ""
```
Integralist, line 5: I noticed that this fixture doesn't specify a `Kind` in its body (which is required, as the code logic will otherwise return an `ErrMissingKind` due to `i.Kind` zero value being zero and so would match `ManagedLoggingUnset`). Yet, the fixture suggests the response was a `200 OK` so maybe this fixture was created before that logic was implemented? Either way I think the fixture might need updating.

Integralist: Actually, I also noticed `ServiceID` is missing from the request body.

Integralist: The tests for the DELETE operation also appear to suggest that Kind/ServiceID is required and they are being validated. So I definitely think the fixture is just out of date.

Integralist (commented):
Thanks @jaredmorrow for opening this PR (and thank you for your patience waiting for a review!). I've a couple of comments/suggested tweaks.

fastly/fixtures/managed_logging/recreate.yaml
jaredmorrow: "Event Buffer" is what is returned from the API currently.

fastly/managed_logging.go
jaredmorrow: Will fix

fastly/fixtures/managed_logging/delete.yaml
jaredmorrow: Pushed a change.  Using `Post` rather than `PostForm`. We don't need Kind or ServiceID in the body.

fastly/fixtures/managed_logging/recreate.yaml
Integralist: OK I guess that might change then by the time of GA maybe? But that's fine 👍🏻

Integralist (approved):

PR #243, opened 2021-01-11
Add region support to logentries logging endpoint

kellymclaughlin:


fastly/logentries_test.go
```diff
@@ -126,6 +130,7 @@ func TestClient_Logentries(t *testing.T) {
 			Name:           "test-logentries",
 			NewName:        String("new-test-logentries"),
 			FormatVersion:  Uint(2),
+			Region:         String("ap"),
```
Integralist, line 133: Do we not need an assertion to validate the update of Region happened successfully?

Integralist (approved):
@kellymclaughlin Left one comment but otherwise LGTM 👍🏻

fastly/logentries_test.go
kellymclaughlin: Good call. I added that check.

PR #245, opened 2021-01-14
Update Test Fixtures

Integralist:
Here are some fixtures I've not updated:

- `managed_logging`: this was only generated in the last few days.
- `custom_tls`, `custom_tls_activation`, `platform_tls`, `tls`: all TLS related fixtures as they required generation of real certificates, asymmetrical keys and other similar resources.
- `kafka`: as the api backend changed behaviour and so I'm going to make that change as a separate PR so it can be explicitly called out in the dynamically updated CHANGELOG.

**NOTES**: this has been a frustrating experience, and one that has made realize we really do need to move away from fixture files for a simpler mocked/dependency injection design.


Integralist:
@christophermancini as discussed on Slack, if you have capacity to review this PR please. Thank you!

fastly/token.go
```diff
@@ -64,17 +64,17 @@ func (c *Client) ListTokens() ([]*Token, error) {
 
 // ListCustomerTokensInput is used as input to the ListCustomerTokens function.
 type ListCustomerTokensInput struct {
-	ID string
+	CustomerID string
```
phamann, line 67: As this is changing functionality, this feels like it should be in a separate PR and not in a test fixtures update PR. As will make documenting in the changelog easier.

fastly/kafka_test.go
```diff
@@ -38,7 +38,7 @@ func TestClient_Kafkas(t *testing.T) {
 			TLSClientCert:    clientCert,
 			TLSClientKey:     clientKey,
 			ParseLogKeyvals:  true,
-			RequestMaxBytes:  1024,
+			RequestMaxBytes:  2048,
```
phamann, line 41: As this is changing functionality, this feels like it should be in a separate PR and not in a test fixtures update PR. As will make documenting in the change-log easier.

phamann (changes requested):
Thanks for doing this. Changes to readme and yaml recordings look good to me. However I feel that the two functionality changes should be in separate PRs and not this one.

fastly/kafka_test.go
Integralist: Changing the value from `1024` to `2048` was a change that _had_ to happen in order to regenerate the fixture 😬 it seems the backend endpoint changed its behaviour.

fastly/token.go
Integralist: 👍🏻

Integralist:
@phamann I've reverted those breaking changes, and I'll move them to a separate PR so they can appear explicitly in the dynamically generated CHANGELOG for a Go-Fastly 3.0.0 release.

> Note: this PR will also be included in that final 3.0.0 release.

/cc @christophermancini

phamann (approved):
👍🏻 LGTM

PR #246, opened 2021-01-18
Cleanup naming of Kinesis to be more consistent.

Integralist:
The naming of our Kinesis public functions are inconsistent and in some cases incorrect. This PR addresses the various references to Kinesis.


phamann (approved):
👍🏻 LGTM

fastly/kinesis.go
```diff
@@ -26,7 +26,7 @@ type Kinesis struct {
 	DeletedAt         *time.Time `mapstructure:"deleted_at"`
 }
 
-// kinesesByName is a sortable list of Kineses.
+// kinesesByName is a sortable list of Kinesis.
```
doramatadora, line 29:
```suggestion
// kinesisByName is a sortable list of Kinesis.
```
And in the type below :)

doramatadora (changes requested):
I think you missed one, otherwise LGTM

fastly/kinesis.go
Integralist: Nice spot!

doramatadora (approved):
image

PR #247, opened 2021-01-19
Avoid generic ID field when more explicit naming would be clearer

Integralist:
There are multiple types of IDs used across our code base and the parent package/function/type doesn't always make it obvious what type of ID is required so we rename a few instances where it would be clearer to have a more explicit name.


phamann (approved):
👍🏻 LGTM

doramatadora (approved):
LGTM 🌮

PR #248, opened 2021-01-19
Update to RequestMaxBytes to align with updated API and regenerate fixtures

Integralist:
The API backend changed behaviour. This meant that the previous _max_ value of `1024` could no longer be used as it was smaller than the allowed _minimum_ of `2048`.


fastly/kafka_test.go
```diff
@@ -38,7 +38,7 @@ func TestClient_Kafkas(t *testing.T) {
 			TLSClientCert:    clientCert,
 			TLSClientKey:     clientKey,
 			ParseLogKeyvals:  true,
-			RequestMaxBytes:  1024,
+			RequestMaxBytes:  2048,
```
phamann, line 41: One minor nit. Seeing as this number is repeated 3 times in this file and an API enforced limit, it feels like it should be factored into a constant value at the top of the file. Just my style preference.

phamann (approved):
👍🏻 LGTM, other than nit about use of constant.

fastly/kafka_test.go
Integralist: I'm usually a "more than twice, refactor" kinda guy. So I can dig it 👍🏻

doramatadora (approved):
I can't comment on the fixtures but the code changes LGTM 🎉

PR #254, opened 2021-01-20
Add important notes related to test fixtures and sensitive data.

Integralist:
In a recent PR where I had to regenerate all our fixtures I realised some aspects of contributing these changes required some additional considerations. I've written up these considerations in the README so future contributors can benefit from them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,6 +186,26 @@ When adding or updating client code and integration tests, contributors should r
 make fix-fixtures FASTLY_TEST_SERVICE_ID="..."
 ```
 
+### Important Test Tips!
+
+There are two important things external contributors need to do when running the tests:
+
+1. Use a 'temporary' token for running the tests (only if regenerating the token fixtures).
+2. Redact sensitive information in your fixtures.
+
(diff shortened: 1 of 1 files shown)


kailan (approved):

README.md
```diff
@@ -186,6 +186,26 @@ When adding or updating client code and integration tests, contributors should r
...
 ```
 
+### Important Test Tips!
+
+There are two important things external contributors need to do when running the tests:
+
+1. Use a 'temporary' token for running the tests (only if regenerating the token fixtures).
+2. Redact sensitive information in your fixtures.
+
+You only need to use a temporary token when regenerating the 'token' fixtures. This is because there is a test to validate the _revoking_ of a token using the [`/tokens/self`](https://developer.fastly.com/reference/api/auth/#revoke-token-current) API endpoint, for which running this test (if there are no existing fixtures) will cause the token you provided at your command-line shell to be revoked/expired. So please don't use a token that's also used by a real/running application! Otherwise you'll discover those application may stop working as you've inadvertently caused your token to be revoked.
```
doramatadora, line 196:
```suggestion
You only need to use a temporary token when regenerating the 'token' fixtures. This is because there is a test to validate the _revoking_ of a token using the [`/tokens/self`](https://developer.fastly.com/reference/api/auth/#revoke-token-current) API endpoint, for which running this test (if there are no existing fixtures) will cause the token you provided at your command-line shell to be revoked/expired. So please don't use a token that's also used by a real/running application! Otherwise you'll discover those applications may stop working as you've inadvertently caused your token to be revoked.
```

PR #255, opened 2021-01-27
Add support for file_max_bytes configuration for azure logging endpoint

kellymclaughlin:
Support file_max_bytes configuration option for azure blob storage logging endpoint.


freeformz:
This is for https://jira.corp.fastly.com/browse/LOG-1229

fastly/blobstorage.go
```diff
@@ -27,6 +27,7 @@ type BlobStorage struct {
 	MessageType       string     `mapstructure:"message_type"`
 	Placement         string     `mapstructure:"placement"`
 	ResponseCondition string     `mapstructure:"response_condition"`
+	FileMaxBytes      uint       `mapstructure:"file_max_bytes"`
```
Integralist, line 30: NOTE: this setting is currently not documented: 
https://developer.fastly.com/reference/api/logging/azureblob/ 

But there is an internal ticket for adding it to the API docs and that ticket is currently 'in review'.

fastly/blobstorage_test.go
```diff
@@ -55,6 +56,7 @@ func TestClient_BlobStorages(t *testing.T) {
 			FormatVersion:   2,
 			MessageType:     "classic",
 			Placement:       "waf_debug",
+			FileMaxBytes:    10 << 20,
```
Integralist, line 59: Can we put these values into constants that can be referenced with a nice descriptive name like `fileBytesExceeded`, `fileBytesAllowed` etc. As that will help make the test code easier to understand 'at a glance'.

**Nitpick/personal preference**: I'd prefer to just use the resulting value rather than a bitwise operator to generate the value, as I now need to open a console and run the left shift to see what the resulting number is (in this case `10485760`). My ability to validate the expectations of the code is made easier when I don't have to do mental hurdle jumps.

Integralist (commented):
@kellymclaughlin LGTM, just have a couple of quick comments.

fastly/blobstorage_test.go
kellymclaughlin: I created a `MiB` constant and updated the `FileMaxBytes` values to be expressed in terms of that constant.

fastly/blobstorage.go
kellymclaughlin: The docs PR is approved and I expect it'll get deployed very soon.

Integralist (approved):

PR #257, opened 2021-02-02
Add more support for compression_codec to logging endpoints

kellymclaughlin:
Add support for compression_codec to the digitalocean and cloudfiles logging endpoints.

This expands the changes from #235 to two more endpoints.


fastly/digitalocean_test.go
```diff
@@ -126,65 +201,108 @@ func TestClient_DigitalOceans(t *testing.T) {
...
 	}
-	if udigitalocean.Name != "new-test-digitalocean" {
-		t.Errorf("bad name: %q", udigitalocean.Name)
+
+	if digitaloceanUpdateResp1.Name != "new-test-digitalocean" {
+		t.Errorf("bad name: %q", digitaloceanUpdateResp1.Name)
+	}
+	if digitaloceanUpdateResp1.Domain != "nyc3.digitaloceanspaces.com" {
+		t.Errorf("bad domain: %q", digitaloceanUpdateResp1.Domain)
+	}
```
Integralist, line 292: Should we not also check:

```go
if digitaloceanUpdateResp1.CompressionCodec != "zstd" {
```

as it was set to snappy originally and now we're setting it to zstd.

fastly/cloudfiles_test.go
```diff
@@ -126,80 +201,121 @@ func TestClient_Cloudfiles(t *testing.T) {
...
+		t.Errorf("bad period: %q", cloudfilesUpdateResp1.Period)
 	}
-	if ucloudfiles.GzipLevel != 0 {
-		t.Errorf("bad gzip_level: %q", ucloudfiles.GzipLevel)
+	if cloudfilesUpdateResp1.FormatVersion != 2 {
+		t.Errorf("bad format_version: %q", cloudfilesUpdateResp1.FormatVersion)
 	}
-	if ucloudfiles.Period != 0 {
-		t.Errorf("bad period: %q", ucloudfiles.Period)
+	if cloudfilesUpdateResp2.CompressionCodec != "zstd" {
```
Integralist, line 307: I mentioned this below in the other test file, but shouldn't we be checking `cloudfilesUpdateResp1` also is set to `zstd` for its compression? as it was `snappy` when initially created.

Integralist (approved):
LGTM, just two small test comments.

fastly/digitalocean_test.go
kellymclaughlin: Good call, I added this check.

fastly/cloudfiles_test.go
kellymclaughlin: Done

Integralist:
Thanks @kellymclaughlin I'll merge this tomorrow and cut a new release in my morning.

PR #259, opened 2021-02-08
Updates needed to support Terraform TLS resources

bengesoff:
There were a mix of tweaks needed while developing the TLS Terraform resources. These range from changes to existing input structs and methods, to some new types and functions for the previously unimplemented functionality.

- Some fields were previously _required_ but they have been changed to _optional_ as per the API documentation
- New methods for specifically [querying TLS Domains](https://developer.fastly.com/reference/api/tls/custom-certs/domains/)
- New methods for managing [TLS Subscriptions](https://developer.fastly.com/reference/api/tls/subs/), excluding the `PATCH` endpoint (coming separately)
- Change pointer usage in Private Keys input struct, and also remove reflection from its filter formatting method.
- Same as above for Bulk Certificate method
- Add `allow_untrusted_root` option to Bulk Certificate creation to allow self-signed CAs to be used
- Update `jsonapi` dependency. I think the previous version was 3 years old and there had been some important changes to allow nested objects in the attributes. This was needed to support parsing Authorization Challenges in the TLS Subscriptions response


fastly/platform_tls.go
```diff
@@ -28,16 +29,19 @@ type TLSConfiguration struct {
 
 // TLSDomain represents a domain (including wildcard domains) that is listed on a certificate's Subject Alternative Names (SAN) list.
 type TLSDomain struct {
-	ID   string `jsonapi:"primary,tls_domain"`
-	Type string `jsonapi:"attr,type"`
+	ID            string                  `jsonapi:"primary,tls_domain"`
+	Type          string                  `jsonapi:"attr,type"`
+	Activations   []*TLSActivation        `jsonapi:"relation,tls_activations,omitempty"`
+	Certificates  []*CustomTLSCertificate `jsonapi:"relation,tls_certificates,omitempty"`
+	Subscriptions []*TLSSubscription      `jsonapi:"relation,tls_subscriptions,omitempty"`
```
Integralist, line 36: I don't see `tls_subscription` in the [TLS domains data model](https://developer.fastly.com/reference/api/tls/custom-certs/domains/) 🤔

fastly/tls_subscription.go
```diff
@@ -0,0 +1,224 @@
...
+	Domains              []*TLSDomain                  `jsonapi:"relation,tls_domains"`
+	Certificates         []*TLSSubscriptionCertificate `jsonapi:"relation,tls_certificates"`
+	Authorizations       []*TLSAuthorizations          `jsonapi:"relation,tls_authorizations"`
+}
+
+type TLSSubscriptionCertificate struct {
+	ID string `jsonapi:"primary,tls_certificate"`
+}
+
+type TLSAuthorizations struct {
```
Integralist, line 30: Am I right in thinking this isn't documented yet (along with TLSChallenge)?

fastly/tls_subscription.go
```diff
@@ -0,0 +1,224 @@
...
+		if !ok {
+			return nil, fmt.Errorf("unexpected response type: %T", data[i])
+		}
+		subscriptions[i] = typed
+	}
+
+	return subscriptions, nil
+}
+
+type CreateTLSSubscriptionInput struct {
```
Integralist, line 119: Could we add in the missing description please.

fastly/tls_subscription.go
```diff
@@ -0,0 +1,224 @@
...
+	var subscription TLSSubscription
+	err = jsonapi.UnmarshalPayload(response.Body, &subscription)
+	if err != nil {
+		return nil, err
+	}
+
+	return &subscription, nil
+}
+
+func domainInSlice(haystack []*TLSDomain, needle *TLSDomain) bool {
```
Integralist, line 154: Could we add in the missing description please. I appreciate it's not publicly exported, but as mentioned earlier ...I'm a stickler for the docs 🙂

fastly/tls_subscription.go
```diff
@@ -0,0 +1,224 @@
...
+	for _, s := range haystack {
+		if s.ID == needle.ID {
+			return true
+		}
+	}
+
+	return false
+}
+
+type GetTLSSubscriptionInput struct {
```
Integralist, line 164: Could we add in the missing description please.

fastly/tls_subscription.go
```diff
@@ -0,0 +1,224 @@
...
+	var subscription TLSSubscription
+	err = jsonapi.UnmarshalPayload(response.Body, &subscription)
+	if err != nil {
+		return nil, err
+	}
+
+	return &subscription, err
+}
+
+type DeleteTLSSubscriptionInput struct {
```
Integralist, line 202: Could we add in the missing description please.

Integralist (commented):
Thanks @bengesoff for your work on this, it's looking great!

Just a couple of minor comments for your consideration.

fastly/platform_tls.go
bengesoff: True but it is in the example response!

fastly/tls_subscription.go
bengesoff: I think so, although I can add some comments regardless if you think that'd be helpful?

Integralist: Yes please 🙂

Integralist (approved):
Thanks @bengesoff this LGTM, with just one minor request to add a missing piece of documentation for `TLSAuthorizations` (I replied to your comment from the last review).

fastly/tls_subscription.go
bengesoff: I've added a comment to this and TLSChallenge -- no official docs to go off of but hopefully it gives an idea what it does 👍🏼

PR #262, opened 2021-02-15
Add PATCH endpoint for TLS Subscriptions

bengesoff:
Left separate from the rest of the TLS Subscriptions updates to enable it to be released independently if needed, as the endpoint is in Limited Availability.

One new method called `UpdateTLSSubscription`. Includes a `Force` flag to allow updates which could potentially break production traffic, although it doesn't actually work without. However, as discussed, this means the user must opt-in to the behaviour and be aware of the risks in doing so.


Integralist (approved):
Thanks @bengesoff LGTM

PR #265, opened 2021-03-24
Use port 443 if UseSSL is set.

Integralist:
**Problem**: in the CLI the `--use-ssl` flag should cause the backend to be created on port 443 (currently no port is set in the call to the go-fastly API client and so the API itself is setting the default of port 80).


Integralist:
Going to close this PR as there are concerns around changing behaviours for backend that aren't ultimately controlled by Fastly. Also this might make code generation of API clients trickier also.

Issue #266, opened 2021-03-25
Purge with Soft=true bug


you-yas55:
Purge with soft=true is broken
```

// Purge instantly purges an individual URL.
func (c *Client) Purge(i *PurgeInput) (*Purge, error) {
	if i.URL == "" {
		return nil, ErrMissingURL
	}

	ro := new(RequestOptions)
	ro.Parallel = true
	if i.Soft {
		ro.Headers["Fastly-Soft-Purge"] = "1"
	}

	resp, err := c.Post("purge/"+i.URL, ro)
	if err != nil {
		return nil, err
	}

	var r *Purge
	if err := decodeBodyMap(resp.Body, &r); err != nil {
		return nil, err
	}
	return r, nil
}

```

this line access a non initialized map maybe! `ro.Headers["Fastly-Soft-Purge"] = "1"`

Integralist:
Thanks for raising this issue @you-yas55 

I'm able to replicate the bug you've described using a test service:

```go
p, err := client.Purge(&fastly.PurgeInput{
	URL:  "http://testing-gofastly-purge.integralist-test.com.global.prod.fastly.net/anything/foo",
	Soft: true,
})
```

Which triggers a panic:

```
panic: assignment to entry in nil map
```

Integralist:
The call to `new(RequestOptions)` only sets a zero value for the `.Headers` map (which for a map type is `nil`) hence trying to assign a value to nil results in a panic.

PR #267, opened 2021-03-25
Initialise Headers map to avoid runtime panic when purging.

Integralist:
**Problem**: https://github.com/fastly/go-fastly/issues/266

The call to `new(RequestOptions)` initialises the struct with zero values. So the `RequestOptions.Headers` map is `nil` by default . Hence trying to assign a value to `nil` results in `panic: assignment to entry in nil map`:

Example code that validated this:

```go
p, err := client.Purge(&fastly.PurgeInput{
	URL:  "http://testing-gofastly-purge.integralist-test.com.global.prod.fastly.net/anything/foo",
	Soft: true,
})
```

### DIFF
--- a/fastly/purge.go
+++ b/fastly/purge.go
@@ -29,6 +29,7 @@ func (c *Client) Purge(i *PurgeInput) (*Purge, error) {
 	ro := new(RequestOptions)
 	ro.Parallel = true
 	if i.Soft {
+		ro.Headers = make(map[string]string)
 		ro.Headers["Fastly-Soft-Purge"] = "1"
 	}
 
--- a/fastly/purge.go
+++ b/fastly/purge.go
@@ -26,11 +26,13 @@ func (c *Client) Purge(i *PurgeInput) (*Purge, error) {
 		return nil, ErrMissingURL
 	}
 
-	ro := new(RequestOptions)
-	ro.Parallel = true
+	ro := &RequestOptions{
+		Parallel: true,
+	}
 	if i.Soft {
-		ro.Headers = make(map[string]string)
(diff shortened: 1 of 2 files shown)


kailan (approved):
Looks good 👍

fastly/purge.go
```diff
@@ -29,6 +29,7 @@ func (c *Client) Purge(i *PurgeInput) (*Purge, error) {
 	ro := new(RequestOptions)
 	ro.Parallel = true
 	if i.Soft {
+		ro.Headers = make(map[string]string)
 		ro.Headers["Fastly-Soft-Purge"] = "1"
 	}
```
phamann, line 34: Small nit: this doesn't follow the conventions throughout the rets of the project or go idioms. Personally i'd write it as:

```suggestion
	ro := &RequestOptions{}
	ro.Parallel = true
	if i.Soft {
		ro.Headers = map[string]string{
			"Fastly-Soft-Purge": 1
		}
	}
```

Integralist: @phamann happy to accept this suggestion, but before I do, regarding your comment...

> or go idioms

I've just read back over the official "Effective Go" and the only distinction they make to using `new` vs manually constructing a struct is when [you have lots of fields to set as part of the initialisation process](https://golang.org/doc/effective_go#composite_literals). Is there a reason why we've opted for `&RequestOptions{}` instead for this simple case (as we don't set any fields)?

phamann: Yes, apologies I missed the `Parallel: true`, definitely prefer you suggestion 👍🏻

PR #269, opened 2021-04-05
Support usage of IAM role in S3 and Kinesis logging endpoints

kellymclaughlin:
Add the ability to specify an IAM role for the S3 and Kinesis logging endpoints.


Integralist (approved):
LGTM. @kellymclaughlin just checking that the developer.fastly.com API reference will be updated as per this change set.

kellymclaughlin:
> LGTM. @kellymclaughlin just checking that the developer.fastly.com API reference will be updated as per this change set.

That's correct. The updates are ready and should be published this week I believe.

PR #271, opened 2021-04-13
Fix testing behavior for logging endpoints the support compression

kellymclaughlin:
Update the testing behavior around the simultaneous use of the `compression_codec` and `gzip_level` options for logging endpoints that support compression.

This relates to discussion on https://github.com/fastly/cli/pull/190.


go.sum
```diff
@@ -34,7 +34,6 @@ golang.org/x/sys v0.0.0-20190215142949-d0b11bdaac8a/go.mod h1:STP8DvDyc/dI5b8T5h
 golang.org/x/sys v0.0.0-20190412213103-97732733099d/go.mod h1:h1NjWce9XRLGQEsW7wpKNCjG9DtNlClVuFLEZdDNbEs=
 golang.org/x/text v0.3.0/go.mod h1:NqM8EUOU14njkJ3fqMW+pc6Ldnwhi/IjpwHt7yyuwOQ=
 golang.org/x/tools v0.0.0-20191119224855-298f0cb1881e/go.mod h1:b+2E5dAYhXwXZwtnZ6UAqBI28+e2cm9otk0dWdXHAEo=
-golang.org/x/tools v0.0.0-20191130070609-6e064ea0cf2d h1:/iIZNFGxc/a7C3yWjGcnboV+Tkc7mxr+p6fDztwoxuM=
```
Integralist, line 37: I presume a `go mod tidy` was run?

Integralist (changes requested):
LGTM. But would be good to not update the fixtures that aren't directly affected by the changed tests.

go.sum
kellymclaughlin: Yep, I remembered to do that this time.

kellymclaughlin:
> LGTM. But would be good to not update the fixtures that aren't directly affected by the changed tests.

I double-checked and I only see changes to fixture files for the affected tests. Was there any in particular you thought didn't belong?

Integralist (approved):
My bad @kellymclaughlin I missed the files at the end of the fixtures 🤦🏻

PR #277, opened 2021-06-07
Add HTTP 206 Partial Content to stats

noseglid:
This is returned by the realtime API, and should be included in the stats struct so it can be used.


Integralist (approved):
Looks good to me. Thanks @noseglid for opening this PR.

noseglid:
When can I expect this to be in a release?

Integralist:
Hi @noseglid 

This would come under a new minor release but I'd ideally have more stuff included in a new release than this one line change 🙂 

I have a bunch of work upcoming for go-fastly, and so I'll likely group this change into that release.

That said, is there an urgency to this being released? Is there a bug you're currently experiencing which currently is tricky to workaround without this change? If so then that might change my perspective on whether the new release should consist this single addition and how quickly.

Thanks.

noseglid:
Nah... I’m missing this stat in my reporting which makes the reports look
kinda odd but nothing too substantial.

I was merely probing for an approximate time but not rushed

Integralist:
👋🏻 @noseglid squeezed a `3.7.0` release here: https://github.com/fastly/go-fastly/releases/tag/v3.7.0

Issue #278, opened 2021-06-07
Support S3 logging `acl`, new `redundancy` options


tpansino:
The SDK for working with S3 logging appears to be missing support for some properties of the [API](https://developer.fastly.com/reference/api/logging/s3/). Specifically:

- `acl` is not currently a supported value
- `redundancy` is missing support for Infrequent Access and One Zone Infrequent Access

[This](https://github.com/fastly/go-fastly/blob/main/fastly/s3.go) looks like the main file that needs to be updated, but also I am not a Go developer. 😅

Integralist:
Hi @tpansino thank you for raising this issue.

API parity with our go-fastly client is something we're actively working on, and so we'll take this and figure out where it sits in the priority order. Unfortunately I can't give an ETA on this, but I'll provide feedback once we're ready to implement.

Integralist:
Hi @tpansino I've released `3.7.0` which has the missing ACL/Redundancy options added:

https://github.com/fastly/go-fastly/releases/tag/v3.7.0

PR #279, opened 2021-06-09
Add S3 Logging ACL parameter and add Redundancy options

bengesoff:
The S3 logging endpoint in the Fastly API supports an `acl` parameter which allows setting the permissions of the uploaded log objects within the bucket.
This commit adds support for this to the SDK, as requested in #278.
The Fastly UI console suggests that there are 7 options for this based on the [AWS "canned ACL"s](https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl), including "private", "public-read" etc, so I have added support for those.

Furthermore, there are 4 options for `redundancy`, but only two were previously supported by the SDK.
I have added the two others for the infrequently accessed storage tiers.

The API docs for s3 logging are here: https://developer.fastly.com/reference/api/logging/s3/

The test fixtures have a lot of changed timestamps and rate limit parameters in the diffs, but the main differences should be the `acl` and `redundancy` parameters in the requests.
I also added a couple of checks to the test to exercise the new options and verify that they were working.


fastly/s3.go
```diff
@@ -9,12 +9,24 @@ import (
...
 	S3ServerSideEncryptionAES S3ServerSideEncryption = "AES256"
 	S3ServerSideEncryptionKMS S3ServerSideEncryption = "aws:kms"
+
+	S3AccessControlListPrivate                S3AccessControlList = "private"
+	S3AccessControlListPublicRead             S3AccessControlList = "public-read"
+	S3AccessControlListPublicReadWrite        S3AccessControlList = "public-read-write"
+	S3AccessControlListAWSExecRead            S3AccessControlList = "aws-exec-read"
+	S3AccessControlListAuthenticatedRead      S3AccessControlList = "authenticated-read"
+	S3AccessControlListBucketOwnerRead        S3AccessControlList = "bucket-owner-read"
+	S3AccessControlListBucketOwnerFullControl S3AccessControlList = "bucket-owner-full-control"
```
Integralist, line 29: There looks to be another value `log-delivery-write` defined in https://docs.aws.amazon.com/AmazonS3/latest/userguide/acl-overview.html#canned-acl.

I presume that extra value was omitted because the Fastly UI didn't make reference to it (which makes sense).

fastly/s3.go
```diff
@@ -9,12 +9,24 @@ import (
...
 type S3ServerSideEncryption string
+type S3AccessControlList string
 
 const (
-	S3RedundancyStandard      S3Redundancy           = "standard"
-	S3RedundancyReduced       S3Redundancy           = "reduced_redundancy"
+	S3RedundancyStandard   S3Redundancy = "standard"
+	S3RedundancyReduced    S3Redundancy = "reduced_redundancy"
+	S3RedundancyOneZoneIA  S3Redundancy = "onezone_ia"
+	S3RedundancyStandardIA S3Redundancy = "standard_ia"
```
Integralist, line 18: There looks to be more values defined in https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html#API_CreateMultipartUpload_RequestParameters (under `x-amz-storage-class`).

I presume those extra values were omitted because the Fastly UI doesn't make reference to them (which makes sense).

fastly/s3.go
```diff
@@ -217,6 +231,7 @@ type UpdateS3Input struct {
 	PublicKey                    *string                `form:"public_key,omitempty"`
 	ServerSideEncryptionKMSKeyID *string                `form:"server_side_encryption_kms_key_id,omitempty"`
 	ServerSideEncryption         S3ServerSideEncryption `form:"server_side_encryption,omitempty"`
+	ACL                          S3AccessControlList    `form:"acl,omitempty"`
```
Integralist, line 234: I think this might need to be a pointer to the type? Although, that said `S3ServerSideEncryption` isn't a pointer 🤔

Integralist (commented):
Thanks @bengesoff I have just one comment that needs some clarification (i.e. the 'struct field as a pointer' comment).

fastly/s3.go
bengesoff: Yes the main reason was because it wasn't mentioned in the UI, but also I guessed that it wasn't relevant anyway as it's a bucket-only policy (as opposed to object-level) and, as far as I'm aware, is only used for S3's own access logs (but I may be wrong)

bengesoff: Also actually if you look at the `x-amz-acl` [request parameter](https://docs.aws.amazon.com/AmazonS3/latest/API/API_CreateMultipartUpload.html#API_CreateMultipartUpload_RequestParameters) then it's not specified in that list either, so I think that confirms the previous comment

fastly/s3.go
bengesoff: Yes similar to the ACL values, the main reason was that they weren't specified in the UI.

fastly/s3.go
bengesoff: I thought that as well, as `S3Redundancy` also isn't a pointer 🤔 I believe the pointer is redundant for all of them actually due to `omitempty` but I just tried to fit in with the existing style - should I change it?

Integralist: Well iirc the reason for using pointers was to allow for checking if something was a zero value or genuinely not provided (because the zero value for a pointer would be nil) and so I feel like it should be changed. In the go-fastly README we even link to a really good article about it: https://willnorris.com/2014/05/go-rest-apis-and-pointers/

bengesoff: Yes you're right - `omitempty` and _not_ having a pointer means you can't `UPDATE` the value to be `""`, I'll change it

fastly/s3_test.go
```diff
@@ -711,7 +711,7 @@ func TestClient_UpdateS3_validation(t *testing.T) {
 		ServiceID:                    "foo",
 		ServiceVersion:               1,
 		Name:                         "test-service",
-		ServerSideEncryption:         S3ServerSideEncryptionKMS,
+		ServerSideEncryption:         S3ServerSideEncryptionPtr(S3ServerSideEncryptionKMS),
```
bengesoff, line 714: @Integralist you'll notice that this is probably the reason most of the `const` fake enums in the SDK don't have pointers in the `PATCH` input structs! Bit of hoop jumping

Integralist: Oh I didn't see the `S3ServerSideEncryptionPtr` thing, so looks like we should replicate that approach.

Integralist: NVM just saw your commit 🙂

Integralist (approved):

Integralist:
@bengesoff did you have any further changes. I'm happy to merge this if you're OK with it being merged.

bengesoff:
@Integralist I'm happy with it if you are!

PR #281, opened 2021-06-23
update VCL snippets

Integralist:
**Problem**: API client logic for VCL snippets didn't align with the API.
**Notes**: The snippet tests have been completely rewritten and new fixtures generated.


fastly/fixtures/vcl_snippets/create_with_required_fields_only.yaml
```diff
@@ -0,0 +1,59 @@
...
+    body: ServiceID=7i6HN3TK9wS159v2gPAZ8A&ServiceVersion=13&content=%23vcl&dynamic=0&name=snipver&type=fetch
+    form:
+      ServiceID:
+      - 7i6HN3TK9wS159v2gPAZ8A
+      ServiceVersion:
+      - "13"
+      content:
+      - '#vcl'
+      dynamic:
+      - "0"
```
triblondon, line 14: Not actually a required field but whatevs :-)

triblondon (approved):

fastly/fixtures/vcl_snippets/create_with_required_fields_only.yaml
Integralist: @triblondon the `dynamic` field is actually required. 

In my test I don't set `dynamic` and so a default value of zero is passed to the API endpoint. 

If I didn't set a default, and just omitted the field altogether in the request to the API endpoint, then the endpoint would respond with...

```json
{"msg":"Bad request","detail":"No value for required field ''dynamic''"}
```

kailan (approved):

fastly/fixtures/vcl_snippets/create_with_required_fields_only.yaml
triblondon: wow, so it is!

PR #290, opened 2021-07-23
ACL entry Subnet should be integer type (missed references)

Integralist:
As per https://github.com/fastly/go-fastly/pull/288 but for the other CRUD methods (i.e. create and single entry update)


fgsch (approved):
🚢 🇮🇹

PR #292, opened 2021-08-17
Add region support for NewRelic logging endpoint

kellymclaughlin:
Add support for the region field to the NewRelic logging endpoint. The valid values for this field are `us` and `eu`. 

The PR to add the API documentation for this is [here](https://github.com/fastly/api-documentation/pull/515).


Integralist (approved):
LGTM.

PR #298, opened 2021-09-13
Add domain validation API

Integralist:
This PR implements the following API endpoints:

- https://developer.fastly.com/reference/api/services/domain/#check-domains
- https://developer.fastly.com/reference/api/services/domain/#check-domain

### DIFF
--- a/fastly/domain.go
+++ b/fastly/domain.go
@@ -318,3 +318,50 @@ type DomainMetadata struct {
 	UpdatedAt *time.Time `json:"updated_at"`
 	DeletedAt *time.Time `json:"deleted_at"`
 }
+
+// ValidateAllDomainsInput is used as input to the ValidateAllDomains function.
+type ValidateAllDomainsInput struct {
+	// ServiceID is the ID of the service (required).
+	ServiceID string
+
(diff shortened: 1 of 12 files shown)


fastly/domain.go
```diff
@@ -222,3 +224,144 @@ func (c *Client) DeleteDomain(i *DeleteDomainInput) error {
...
+// DomainValidationResult defines an idiomatic representation of the API
+// response.
+type DomainValidationResult struct {
+	Metadata DomainMetadata
+	CName    string
+	Valid    bool
+}
+
+// UnmarshalJSON works around the badly designed API response by coercing the
+// raw data into a more appropriate data structure.
```
Integralist, line 284: @kailan @doramatadora @triblondon 

This is an interesting example of where code generating the go-fastly client might present problems. 

e.g. it's unlikely that the generated code will try to handle this situation (i.e. multiple types within an array) intelligently as it won't know the context of the response data. 

I imagine it will more likely just generate code that uses the empty interface type `interface{}`, which will cause more problems for consumers of the library as they'd then need to type assert the `interface{}` into a concrete type.

triblondon (approved):

PR #309, opened 2021-10-04
Add `UseTLS` field for Splunk logging

smaeda-ks:
and I'll be adding support in our TF provider as well once a new version is released:
https://registry.terraform.io/providers/fastly/fastly/latest/docs/resources/service_v1#nested-schema-for-splunk

main diff: https://github.com/fastly/go-fastly/pull/309/commits/3bbbf3751efbd355fe5286979653d600ebdc810a


smaeda-ks:
We probably should update:
https://github.com/fastly/go-fastly/blob/f1303d97fa9a186aa1b15d246a96982362dceed4/fastly/client.go#L49-L50

Integralist:
@smaeda-ks doh! yes we should update that to 5.0.0.

Can you do that in this PR please. Thanks.

Integralist (approved):
LGTM. If the ProjectVersion field could also be bumped (appreciate it's not related but might as well just get that change merged ASAP).

smaeda-ks:
@Integralist Thank you. Added a commit to bump the `ProjectVersion`

PR #315, opened 2021-11-02
Replace form dependency with google's go-querystring

Integralist:
Fixes https://github.com/fastly/terraform-provider-fastly/issues/495

The big change is in [`fastly/client.go`](https://github.com/fastly/go-fastly/pull/315/files#diff-0a6526414abb97857aad37b9c7cf08a1082724a577d2b38d9c04c9da131aae7f) and in [`fastly/token.go`](https://github.com/fastly/go-fastly/pull/315/files#diff-11702f23f29d1dc70e860144f2215d230aecc26d41c8531a4c40cbaad2387e19).

All other changes are in updating the struct tags from `form` to `url` which is what the new dependency uses.

You'll also notice in [`fastly/vcl_test.go`](https://github.com/fastly/go-fastly/pull/315/files#diff-6e7fb15c6f86a1bbbc0eaf67976ca198d987402390586447040edba33235386a) I've added some extra VCL code that uses a regex with a pipe operator so that I can be sure nothing odd happens with the VCL.


fastly/token.go
```diff
@@ -107,12 +107,12 @@ func (c *Client) GetTokenSelf() (*Token, error) {
...
-	Scope     TokenScope `form:"scope,omitempty"`
-	Username  string     `form:"username,omitempty"`
-	Password  string     `form:"password,omitempty"`
-	Services  []string   `form:"services,omitempty"`
-	ExpiresAt *time.Time `form:"expires_at,omitempty"`
+	Name      string     `url:"name,omitempty"`
+	Scope     TokenScope `url:"scope,omitempty"`
+	Username  string     `url:"username,omitempty"`
+	Password  string     `url:"password,omitempty"`
+	Services  []string   `url:"services,brackets,omitempty"`
```
smaeda-ks, line 114: note for others

> Including the "brackets" option signals that the multiple URL values should have "[]" appended to the value name. 

https://pkg.go.dev/github.com/google/go-querystring/query

smaeda-ks:
Looks like this PR breaks `Compatibool` functionality:
https://github.com/fastly/go-fastly/blob/main/fastly/fastly.go

We're currently relying on the custom marshaling feature of the `ajg/form` package to support `Compatibool`:
https://github.com/ajg/form#custom-marshaling

Integralist:
@triblondon @smaeda-ks I've attempted to update as many fixtures as possible but there are some I've not because of complications that I don't have capacity for at the moment.

Shohei has run this code against the terraform test suite and it's passing so we have confidence without the need to have 100% updated fixtures.

**Fixtures not updated**:

- Any of the TLS fixtures: This is a complex set of configuration that needs to be in place and so I'm not going to waste time trying to recreate it.
- Token fixtures: I remember it was pretty tedious messing around with the tokens so I've decided to just side-step it for now.
- `TestClient_EdgeCheck`: 403 Forbidden "You are not authorized to perform this action" (the test tries to validate against `releases.hashicorp.com`).
- `TestClient_Package`: 400 - Bad Request. Not sure why this failed as the test is trying to update the service with an invalid package. I didn't want to stop my flow to investigate.

smaeda-ks (approved):

PR #318, opened 2021-11-12
Do not omit zero int values on create operations

smaeda-ks:
(mainly for TF provider)

https://github.com/fastly/terraform-provider-fastly/issues/129
https://github.com/fastly/terraform-provider-fastly/issues/109
https://github.com/fastly/terraform-provider-fastly/issues/87

### DIFF
--- a/fastly/backend.go
+++ b/fastly/backend.go
@@ -98,15 +98,15 @@ type CreateBackendInput struct {
 	Name                string      `url:"name,omitempty"`
 	Comment             string      `url:"comment,omitempty"`
 	Address             string      `url:"address,omitempty"`
-	Port                uint        `url:"port,omitempty"`
+	Port                *uint       `url:"port,omitempty"`
 	OverrideHost        string      `url:"override_host,omitempty"`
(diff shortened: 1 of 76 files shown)


smaeda-ks:
@Integralist Do you think this requires updating test fixtures?

smaeda-ks:
Also, we seem to have some inconsistencies between `int` vs `uint`.

Integralist:
I'd personally prefer to regenerate the fixtures.

I agree there is an inconsistency. I typically use `int` by default, but I can appreciate that if we know we'll only ever need positive number then using `uint` makes sense. I wouldn't worry too much about this for now, especially if we consider in the future moving go-fastly to being code-generated.

Integralist (approved):
Approving ahead of fixtures being regenerated.

smaeda-ks:
@Integralist updated fixtures.

Integralist (approved):

smaeda-ks:
ok to merge? I don't have write access though

Integralist:
Oh, sorry @smaeda-ks I'll merge now 😬

PR #321, opened 2021-12-09
Add support for Link header pagination

smaeda-ks:
client example:

```go
	paginator := client.NewListACLEntriesPaginator(
		&fastly.ListACLEntriesInput{
			ServiceID: serviceID,
			ACLID:     "your_acl_id",
			PerPage:   50,
		},
	)

	var es []*fastly.ACLEntry
	for paginator.HasNext() {
		data, err := paginator.GetNext()
		if err != nil {
			break
		}
		es = append(es, data...)
	}
```

Adding `NewListACLEntriesPaginator` and `NewListDictionaryItemsPaginator`.


smaeda-ks:
@Integralist let me know if you have any questions.

fastly/acl_entry.go
```diff
@@ -63,6 +71,112 @@ func (c *Client) ListACLEntries(i *ListACLEntriesInput) ([]*ACLEntry, error) {
...
+
+	if i.Direction != "" {
+		requestOptions.Params["direction"] = i.Direction
+	}
+	if i.Sort != "" {
+		requestOptions.Params["sort"] = i.Sort
+	}
+
+	resp, err := c.Get(path, requestOptions)
+	if err == nil {
```
Integralist, line 150: I think we should replace the if/else with a single if conditional (as it aligns more closely with the traditional error handling pattern and avoids needing an `else` block)...

```go
if err != nil {
    return nil, err
}

for _, l := range link.ParseResponse(resp) {
    // ...
}

p.consumed = true
```

fastly/acl_entry.go
```diff
@@ -63,6 +71,112 @@ func (c *Client) ListACLEntries(i *ListACLEntriesInput) ([]*ACLEntry, error) {
...
+		return nil, ErrMissingACLID
+	}
+
+	var perPage int
+	if i.PerPage <= 0 {
+		// Max is 100
+		perPage = 100
+	} else {
+		perPage = i.PerPage
+	}
```
Integralist, line 126: Maybe refactor this code to...

```go
const maxPerPage = 100 

perPage := i.PerPage
if perPage <= 0 {
    perPage = maxPerPage
}
```

fastly/dictionary_item.go
```diff
@@ -63,6 +69,110 @@ func (c *Client) ListDictionaryItems(i *ListDictionaryItemsInput) ([]*Dictionary
...
+		return nil, ErrMissingDictionaryID
+	}
+
+	var perPage int
+	if i.PerPage <= 0 {
+		// Max is 100
+		perPage = 100
+	} else {
+		perPage = i.PerPage
+	}
```
Integralist, line 123: Maybe refactor this code to...

```go
const maxPerPage = 100 

perPage := i.PerPage
if perPage <= 0 {
    perPage = maxPerPage
}
```

Integralist (approved):
LGTM but have a couple of suggestions.

Out of interest, do you happen to know how to get the [generated documentation](https://pkg.go.dev/github.com/fastly/go-fastly/v5/fastly#Client.ListACLEntries) to display an example snippet like in your PR description (not sure if that's done by adding the snippet into a code comment above the code function)?

smaeda-ks:
@Integralist Thanks. Applied all suggestions and added code examples. Tested this locally using godoc. This should work.
Screen Shot 2022-01-06 at 3 44 48

Integralist (approved):

PR #322, opened 2022-01-04
Add support for `modsec_rule_id` filter parameter

smaeda-ks:
https://developer.fastly.com/reference/api/waf/rules/#list-waf-rules

This is to support https://github.com/fastly/terraform-provider-fastly/issues/520

### DIFF
--- a/fastly/waf_rules.go
+++ b/fastly/waf_rules.go
@@ -49,6 +49,8 @@ type ListWAFRulesInput struct {
 	FilterTagNames []string
 	// Limit the returned rules to a set by publishers.
 	FilterPublishers []string
+	// Limit the returned rules to a set by modsecurity rule IDs.
+	FilterModSecIDs []int
 	// Excludes individual rules by modsecurity rule IDs.
 	ExcludeMocSecIDs []int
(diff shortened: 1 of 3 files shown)


smaeda-ks:
@Integralist

fastly/waf_rules.go
```diff
@@ -49,6 +49,8 @@ type ListWAFRulesInput struct {
 	FilterTagNames []string
 	// Limit the returned rules to a set by publishers.
 	FilterPublishers []string
+	// Limit the returned rules to a set by modsecurity rule IDs.
+	FilterModSecIDs []int
 	// Excludes individual rules by modsecurity rule IDs.
 	ExcludeMocSecIDs []int
```
Integralist, line 55: Out if interest this looks like a typo... right? Should be `ExcludeModSecIDs`? If so we should add a TODO to update and bump to a new major version.

Integralist (approved):
LGTM. Can we regenerate the fixtures to help validate the changes.

smaeda-ks:
@Integralist Thanks. Added a test for this change.

PR #325, opened 2022-01-10
Update Go modules

smaeda-ks:
Just trying to update modules...

I'm updating tests in `custom_tls_certificate_test` as I immediately noticed that their fixtures are pretty old and the response data format is different than what we return today, and it was causing errors after updating the `jsonapi` module:

```sh
$ make all
==> Downloading Go module
==> Downloading development dependencies
==> Tidying module
==> Running gofmt
==> Fixing imports
==> Testing go-fastly
--- FAIL: TestClient_ListCustomTLSCertificates_validation (0.00s)
    custom_tls_certificate_test.go:145: Trying to Unmarshal an object of type "", but "tls_certificate" does not match
--- FAIL: TestClient_CustomTLSCertificate (0.00s)
    custom_tls_certificate_test.go:40: Trying to Unmarshal an object of type "", but "tls_certificate" does not match
```

https://github.com/fastly/go-fastly/commit/b71bd99c2f5b9651bda377c740bd6d2a1fae6f88#diff-c0c259c2589ed26c6456d3f1a503dfad3bbf0d45a28b5ff913b90a86a687df6cL15-R15

you notice that the old fixture had an extra `[{"data":{` block.


smaeda-ks:
cc: @Integralist

Integralist (approved):

PR #328, opened 2022-01-12
Pagination to return interface not concrete type

Integralist:
The recent pagination implementations were designed to return concrete types, which include methods attached to the returned type that in some cases are responsible for making API calls.

In the Fastly CLI we're unable to mock these returned concrete types and so in this PR I've swapped them out for interfaces.

> **NOTE**: We can reduce the boilerplate once go 1.18 comes out and we have access to generics.

This is a breaking change and will require a new major version of the go-fastly client.

### DIFF
--- a/fastly/dictionary_item.go
+++ b/fastly/dictionary_item.go
@@ -97,7 +97,7 @@ func (p *ListDictionaryItemsPaginator) GetNext() ([]*DictionaryItem, error) {
 }
 
 // NewListDictionaryItemsPaginator returns a new paginator
-func (c *Client) NewListDictionaryItemsPaginator(i *ListDictionaryItemsInput) PaginatorDictItems {
+func (c *Client) NewListDictionaryItemsPaginator(i *ListDictionaryItemsInput) PaginatorDictionaryItems {
 	return &ListDictionaryItemsPaginator{
 		client:  c,
 		options: i,
--- a/fastly/dictionary_item.go
+++ b/fastly/dictionary_item.go
@@ -96,8 +96,8 @@ func (p *ListDictionaryItemsPaginator) GetNext() ([]*DictionaryItem, error) {
 	return p.client.listDictionaryItemsWithPage(p.options, p)
 }
 
-// NewListDictionaryItemsPaginator returns a new ListDictionaryItemsPaginator
-func (c *Client) NewListDictionaryItemsPaginator(i *ListDictionaryItemsInput) *ListDictionaryItemsPaginator {
+// NewListDictionaryItemsPaginator returns a new paginator
+func (c *Client) NewListDictionaryItemsPaginator(i *ListDictionaryItemsInput) PaginatorDictItems {
 	return &ListDictionaryItemsPaginator{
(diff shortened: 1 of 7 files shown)


fastly/dictionary_item.go
```diff
@@ -96,8 +96,8 @@ func (p *ListDictionaryItemsPaginator) GetNext() ([]*DictionaryItem, error) {
 	return p.client.listDictionaryItemsWithPage(p.options, p)
 }
 
-// NewListDictionaryItemsPaginator returns a new ListDictionaryItemsPaginator
-func (c *Client) NewListDictionaryItemsPaginator(i *ListDictionaryItemsInput) *ListDictionaryItemsPaginator {
+// NewListDictionaryItemsPaginator returns a new paginator
+func (c *Client) NewListDictionaryItemsPaginator(i *ListDictionaryItemsInput) PaginatorDictItems {
```
smaeda-ks, line 100:
```suggestion
func (c *Client) NewListDictionaryItemsPaginator(i *ListDictionaryItemsInput) PaginatorDictionaryItems {
```

fastly/dictionary_item_test.go
```diff
@@ -67,7 +67,7 @@ func TestClient_DictionaryItems(t *testing.T) {
 
 	// List with paginator
 	var dictionaryItems2 []*DictionaryItem
-	var paginator *ListDictionaryItemsPaginator
+	var paginator PaginatorDictItems
```
smaeda-ks, line 70:
```suggestion
	var paginator PaginatorDictionaryItems
```

fastly/paginator.go
```diff
@@ -0,0 +1,24 @@
...
+
+// PaginatorACLEntries represents a paginator.
+type PaginatorACLEntries interface {
+	HasNext() bool
+	Remaining() int
+	GetNext() ([]*ACLEntry, error)
+}
+
+// PaginatorDictItems represents a paginator.
+type PaginatorDictItems interface {
```
smaeda-ks, line 13:
```suggestion
// PaginatorDictionaryItems represents a paginator.
type PaginatorDictionaryItems interface {
```

smaeda-ks (changes requested):
LGTM, except for some naming convention consistencies suggestions I just left. And if this is going to be a major version release it's probably better to include https://github.com/fastly/go-fastly/pull/325 as well.

Integralist:
@smaeda-ks OK cool thanks. I'll merge this and I'll look at #325 (do you know if that's almost ready for a review)?

smaeda-ks:
@Integralist As for #325, note quite yet. Due to how go-fastly runs tests using recorded fixtures, I'm still not 100% confident to say it's okay to merge as it's not making real requests each time. So my idea is to run TF provider full tests using that branch and see if anything comes up. Since our provider is a large consumer of this library we could rely on that result relatively confidently. Thoughts?

Integralist:
@smaeda-ks SGTM 👍🏻 OK let me know when it's ready for a review, and or if you need any help.

PR #333, opened 2022-02-14
Stats: Add test for fetching stats by service and field

hermanschaaf:
Before this change, there was no test covering the `/stats/service/:service_id/field/:field` endpoint ([docs](https://developer.fastly.com/reference/api/metrics-stats/historical-stats/#get-hist-stats-service-field)). This endpoint gets used when both a service ID and field name are given in the options.

The surprising thing about this case is that despite being for a specific field, when specifying both a service and field, one needs to be call the `GetStats` method, not the `GetStatsField` method, due to the format of the response. The new test case provides an example of proper usage.

### DIFF
--- a/fastly/fixtures/stats/service_stats_by_field_and_service.yaml
+++ b/fastly/fixtures/stats/service_stats_by_field_and_service.yaml
@@ -1,0 +1,40 @@
+---
+version: 1
+interactions:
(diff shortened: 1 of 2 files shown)


Integralist (approved):
Thanks for adding this test case 🙂

Issue #336, opened 2022-02-15
api throttling not exposed


dkegel-fastly:
The fastly api provides the http response header Fastly-RateLimit-Remaining
( https://developer.fastly.com/reference/api/ ), which can be used to monitor
how well an app is dealing with api throttling, or as an input to the app's own api call rate limiting.
go-fastly does not currently return this information to the caller.

I will submit a proof of concept pull request showing one low-impact way of exposing this information to apps which need it.

PR #337, opened 2022-02-15
Support `Fastly-RateLimit-Remaining`

dkegel-fastly:
Lets apps observe how close they are to a `429 Too Many Requests` HTTP response from Fastly.

https://developer.fastly.com/reference/api/#rate-limiting

Fixes https://github.com/fastly/go-fastly/issues/336

### DIFF
--- a/fastly/client.go
+++ b/fastly/client.go
@@ -14,6 +14,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"time"
 
 	"github.com/google/go-querystring/query"
 	"github.com/google/jsonapi"
@@ -74,6 +75,9 @@ type Client struct {
 
 	// remaining is last observed value of http header Fastly-RateLimit-Remaining
 	remaining int
+
+	// reset is last observed value of http header Fastly-RateLimit-Reset
+	reset int64
 }
 
 // RTSClient is the entrypoint to the Fastly's Realtime Stats API.
@@ -148,6 +152,7 @@ func (c *Client) init() (*Client, error) {
 	// Use the default limit as a first guess:
 	// https://developer.fastly.com/reference/api/#rate-limiting
 	c.remaining = 1000
+	// Remaining() is left at the epoch to indicate a refresh of remaining is overdue
 
 	u, err := url.Parse(c.Address)
 	if err != nil {
@@ -166,6 +171,11 @@ func (c *Client) init() (*Client, error) {
 // rate limiting causes a 429 Too Many Requests error.
 func (c *Client) Remaining() int {
 	return c.remaining
+}
+
+// Reset returns the next time the rate limiter's Remaining counter will be refilled.
+func (c *Client) Reset() time.Time {
+	return time.Unix(c.reset, 0)
 }
 
 // Get issues an HTTP GET request.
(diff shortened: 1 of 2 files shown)


dkegel-fastly:
Some users will also want to see the Fastly-RateLimit-Reset header.

An alternate approach might be to register a callback that gets both headers whenever they are received.

fastly/client.go
```diff
@@ -275,6 +290,16 @@ func (c *Client) Request(verb, p string, ro *RequestOptions) (*http.Response, er
 		return resp, err
 	}
 
+	if verb != "GET" && verb != "HEAD" {
+		remaining := resp.Header.Get("Fastly-RateLimit-Remaining")
+		if remaining != "" {
+			val, err := strconv.Atoi(remaining)
+			if err == nil {
+				c.remaining = val
+			}
```
Integralist, line 299: The `val` variable isn't used after this block so might as well inline the conditional...

```suggestion
			if val, err := strconv.Atoi(remaining); err == nil {
				c.remaining = val
			}
```

dkegel-fastly: Done.

Integralist (commented):
LGTM but can we also add an example file ([example](https://github.com/fastly/go-fastly/blob/72704ae27093ad4f1a3d6159d7a6ad372e28e7cb/fastly/example_service_test.go)) so users who are looking through the `pkg.go.dev` documentation can better understand how they might use the new `Remaining()` method. I appreciate the use case is fairly straightforward but worth having an example I think.

Integralist:
Oh, and maybe it's worth implementing a similar solution for `Fastly-RateLimit-Reset`.

dkegel-fastly:
The method name is a bit awkward...

fastly/client.go
```diff
@@ -153,6 +167,17 @@ func (c *Client) init() (*Client, error) {
...
 }
 
+// Remaining returns the number of non-read requests left before
+// rate limiting causes a 429 Too Many Requests error.
+func (c *Client) Remaining() int {
+	return c.remaining
+}
+
+// Reset returns the next time the rate limiter's Remaining counter will be refilled.
+func (c *Client) Reset() time.Time {
```
Integralist, line 177: Maybe we rename to `RateLimitRemaining` and `RateLimitReset`.

dkegel-fastly: Done.

fastly/client.go
```diff
@@ -153,6 +167,17 @@ func (c *Client) init() (*Client, error) {
 	return c, nil
 }
 
+// RateLimitRemaining returns the number of non-read requests left before
+// rate limiting causes a 429 Too Many Requests error.
+func (c *Client) RateLimitRemaining() int {
+	return c.remaining
+}
+
+// Reset returns the next time the rate limiter's counter will be refilled.
```
Integralist, line 176:
```suggestion
// RateLimitReset returns the next time the rate limiter's counter will be
// reset.
```

fastly/client.go
```diff
@@ -140,6 +148,12 @@ func NewRealtimeStatsClientForEndpoint(token, endpoint string) (*RTSClient, erro
 }
 
 func (c *Client) init() (*Client, error) {
+	// Until we do a request, we don't know how many are left.
+	// Use the default limit as a first guess:
+	// https://developer.fastly.com/reference/api/#rate-limiting
+	c.remaining = 1000
+	// c.reset is left at the epoch to indicate a refresh of remaining is overdue
```
Integralist, line 155:
```suggestion
```

Integralist (approved):

PR #340, opened 2022-03-01
go.mod: downgrade to 1.16

Integralist:
The upgrade to 1.17 caused problems with some customers and internal teams because of changes in the language that caused (as an example) unexpected panics to be triggered. 

No features from 1.17 are being used, we bumped the version to be on the latest and not expecting there to be any issue.

> **NOTE**: After discussing internally, it's likely we'll bump to 1.17 when 1.19 is released (so as to stay two versions behind the latest -- yes 1.18 isn't released yet but we're close 😉).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,11 +14,4 @@ require (
 	honnef.co/go/tools v0.2.2
 )
 
-require (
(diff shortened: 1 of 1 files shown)


dkegel-fastly (approved):
LGTM

PR #349, opened 2022-06-15
Support for Service Authorizations

noseglid:
When a user has the "engineer" level, specific access to specific
services can be granted using service authorizations.

This change adds support for creating, retrieving, updating and deleting
service authorizations using the Go client.

### DIFF
--- a/fastly/errors.go
+++ b/fastly/errors.go
@@ -171,6 +171,14 @@ var ErrMissingServerSideEncryptionKMSKeyID = NewFieldError("ServerSideEncryption
 // requires a "ServiceID" key, but one was not set.
 var ErrMissingServiceID = NewFieldError("ServiceID")
 
+// ErrMissingUserID is an error that is returned when an input struct
+// requires a "UserID" key, but one was not set
+var ErrMissingUserID = NewFieldError("UserID")
+
+// ErrMissingPermissions is an error that is returned when an input struct
+// requires a "Permissions" key, but one was not set
+var ErrMissingPermissions = NewFieldError("Permissions")
+
 // ErrMissingServiceVersion is an error that is returned when an input struct
 // requires a "ServiceVersion" key, but one was not set.
 var ErrMissingServiceVersion = NewFieldError("ServiceVersion")
--- a/fastly/errors.go
+++ b/fastly/errors.go
@@ -176,7 +176,7 @@ var ErrMissingServiceID = NewFieldError("ServiceID")
 var ErrMissingServiceAuthorizationsService = NewFieldError("Service").Message("SAService requires an ID")
 
 // ErrMissingServiceAuthorizationsUser is an error that is returned when an input struct
-// requires a "ServiceAuthorizationUser" key, but one was not set.
+// requires a "User" key of type SAUser, but one was not set or was misconfigured.
 var ErrMissingServiceAuthorizationsUser = NewFieldError("User").Message("SAUser requires an ID")
 
 // ErrMissingUserID is an error that is returned when an input struct
--- a/fastly/errors.go
+++ b/fastly/errors.go
@@ -172,7 +172,7 @@ var ErrMissingServerSideEncryptionKMSKeyID = NewFieldError("ServerSideEncryption
 var ErrMissingServiceID = NewFieldError("ServiceID")
 
 // ErrMissingServiceAuthorizationsService is an error that is returned when an input struct
-// requires a "ServiceAuthorizationService" key, but one was not set.
+// requires a "Service" key of type SAService, but one was not set or was misconfigured.
 var ErrMissingServiceAuthorizationsService = NewFieldError("Service").Message("SAService requires an ID")
 
 // ErrMissingServiceAuthorizationsUser is an error that is returned when an input struct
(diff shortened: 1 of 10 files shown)


Integralist (changes requested):
Hi @noseglid 👋🏻 

Thank you for opening this PR. It's really appreciated.

For the sake of consistency with the rest of the code base I've requested some changes to help keep the code aligned.

fastly/service_authorization.go
```diff
@@ -0,0 +1,202 @@
+package fastly
+
+import (
+	"fmt"
+)
+
+type ServiceAuthorizationPermission string
+type ServiceAuthorizationType string
```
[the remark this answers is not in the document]
noseglid: Makes sense, I'll remove them.

Regarding the Terraform change, we get away from the type conversion, but not the type assertion as  the `Get` func returns `interface{}`. The referenced line would instead be
```go
d.Get("permission").(string)
```
which I agree is better 👍

fastly/service_authorization_test.go
```diff
@@ -0,0 +1,146 @@
...
+	t.Parallel()
+
+	fixtureBase := "service_authorizations/"
+
+	// Create
+	var err error
+	var sa *ServiceAuthorization
+	record(t, fixtureBase+"create", func(c *Client) {
+		sa, err = c.CreateServiceAuthorization(&CreateServiceAuthorizationInput{
+			ServiceID:  "7ZVxm5pPWdzKdl3P5UW7jR",
```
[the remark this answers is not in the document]
noseglid: I see. Definitely makes sense. Thanks

fastly/service_authorization.go
```diff
@@ -0,0 +1,127 @@
...
+	"github.com/google/jsonapi"
+)
+
+type SAUser struct {
+	ID string `jsonapi:"primary,user"`
+}
+
+type SAService struct {
+	ID string `jsonapi:"primary,service"`
+}
```
noseglid, line 16: It would be great if the `User` struct already was `jsonapi`'d and I could re-use it directly. But the way it currently stands, the `User` struct still has the `mapstructure` tags which makes it incompatible with the `jsonapi` exposed by service authorizations (if I understood it correctly).

Alternatively to creating these structs I could add the relevant `jsonapi` to the [User](https://github.com/fastly/go-fastly/blob/main/fastly/user.go#L12) and [Service](https://github.com/fastly/go-fastly/blob/main/fastly/service.go#L15) structs and re-use them, but I'm not sure mixing tags like this is the best course. Thoughts?

Integralist: Yeah I would tend to agree and opt for not overlapping the tags just in case it causes any unexpected issues. 👍🏻 

So I think recreating the structs is fine.

fastly/service_authorization.go
```diff
@@ -0,0 +1,140 @@
...
+	User *SAUser `jsonapi:"relation,user,omitempty"`
+}
+
+// CreateServiceAuthorization creates a new service authorization granting granular service and user permissions.
+func (c *Client) CreateServiceAuthorization(i *CreateServiceAuthorizationInput) (*ServiceAuthorization, error) {
+	if i.Service == nil || i.Service.ID == "" {
+		return nil, ErrMissingServiceID
+	}
+	if i.User == nil || i.User.ID == "" {
+		return nil, ErrMissingUserID
```
Integralist, line 75: Similar comment to above:

```go
var ErrMissingServiceAuthorizationsUser = NewFieldError("User").Message("SAUser requires an ID")
```

fastly/fastly_test.go
```diff
@@ -19,9 +19,15 @@ var testStatsClient = NewRealtimeStatsClient()
...
 var testServiceID = serviceIDForTest()
 
+// testUserID is the ID of the testing user.
+var testUserID = userIDForTest()
+
 // Default ID of the testing service.
 var defaultTestServiceID = "7i6HN3TK9wS159v2gPAZ8A"
 
+// Default ID of the testing user.
+var defaultTestUserID = "4tKBSuFhNEiIpNDxmmVydt"
```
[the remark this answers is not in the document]
Integralist: And remove associated changes elsewhere like in `scripts/fixFixtures.sh`.

Integralist (commented):
Thanks @noseglid this is looking a lot better. Just a few small comments. Thanks!

noseglid:
I have now addressed all your comments

fastly/errors.go
```diff
@@ -171,6 +171,22 @@ var ErrMissingServerSideEncryptionKMSKeyID = NewFieldError("ServerSideEncryption
 // requires a "ServiceID" key, but one was not set.
 var ErrMissingServiceID = NewFieldError("ServiceID")
 
+// ErrMissingServiceAuthorizationsService is an error that is returned when an input struct
+// requires a "ServiceAuthorizationService" key, but one was not set.
```
Integralist, line 175:
```suggestion
// requires a "Service" key of type SAService, but one was not set or was misconfigured.
```

fastly/errors.go
```diff
@@ -171,6 +171,22 @@ var ErrMissingServerSideEncryptionKMSKeyID = NewFieldError("ServerSideEncryption
...
+var ErrMissingServiceAuthorizationsService = NewFieldError("Service").Message("SAService requires an ID")
+
+// ErrMissingServiceAuthorizationsUser is an error that is returned when an input struct
+// requires a "ServiceAuthorizationUser" key, but one was not set.
```
Integralist, line 179:
```suggestion
// requires a "User" key of type SAUser, but one was not set or was misconfigured.
```

Integralist (approved):
Made two suggested edits (which I'll apply in a moment), but otherwise this looks good to me.

Thanks @noseglid 🙂

noseglid:
🙌

PR #359, opened 2022-08-23
add object store api support

dgryski:


fastly/object_store.go
```diff
@@ -0,0 +1,214 @@
...
+import (
+	"io"
+	"net/http"
+	"strings"
+	"time"
+)
+
+// https://developer.fastly.com/reference/api/object-store/
+
+// TODO(dgryski): Should the store id be `StoreID` or `ID` consistently?
```
Integralist, line 12: It's historically been mixed, but I tend to use `ID` and rely on the name of the struct to provide the context for the type of ID. If it's ambiguous or there is more than one type of ID that can be passed to the struct then I'd lean towards a more explicit `StoreID`.

dgryski: Ok, then I can reuse the `ErrMissingID` error instead of creating a new one.  I'll update field to be just `ID`.

fastly/object_store.go
```diff
@@ -0,0 +1,214 @@
...
+}
+
+type ListObjectStoresResponse struct {
+	Data []ObjectStore
+	Meta map[string]string
+}
+
+func (c *Client) ListObjectStores() (*ListObjectStoresResponse, error) {
+	const path = "/resources/stores/object"
+	// TODO(dgryski): How do I handle pagination?
```
Integralist, line 50: I'd say look to https://github.com/fastly/go-fastly/blob/0092ed3705eca1fe2bbcca43a863c9d49471e626/fastly/dictionary_item.go#L72 for an example implementation.

Original PR: https://github.com/fastly/go-fastly/pull/321

Integralist: iirc I had suggested maybe refactoring the paginator once we got to go 1.18 because having to recreate the paginator logic across types could be tedious, but A.) go-fastly won't support 1.18 anytime soon, and B.) even if it did I can't recall if it's possible to abstract with generics.

dgryski:
We seriously test against Go 1.14?

Integralist:
> We seriously test against Go 1.14?

I think that's just something that we've never updated. So feel free to update the CI to use at least 1.16

dgryski:
@Integralist ping

fastly/object_store_test.go
```diff
@@ -0,0 +1,155 @@
...
+		sort.Strings(listed)
+		if !reflect.DeepEqual(keys, listed) {
+			t.Errorf("mismatch listing paginated keys: got %q, want %q", objectStoreListKeys.Data, keys)
+		}
+	})
+
+	// create more stores
+	// list
+	// list with pagination
+	// cleanup
```
Integralist, line 153: Delete?
```suggestion
```

fastly/object_store.go
```diff
@@ -0,0 +1,419 @@
...
+func (l *ListObjectsStoreKeysPaginator) Err() error {
+	return l.err
+}
+
+// Keys returns the current set of keys retrieved by the paginator.
+func (l *ListObjectsStoreKeysPaginator) Keys() []string {
+	return l.keys
+}
+
+// missing example requests from API reference
```
Integralist, line 322: Delete?
```suggestion
```

Integralist (approved):
Couple of suggested edits.

dgryski:
Old comments removed.

Integralist:
Cool thanks 👍🏻 

Feel free to go ahead and merge (or do you need me to)?

chenrui333:
Any idea when this would be supported in the provider side?

PR #363, opened 2022-09-07
Ensure http.Response.Body is closed

### LINKED ISSUE #362: HTTP client never closes response body
This leads to `too many open files` error. 

[According](https://pkg.go.dev/net/http) to the `net/http` package docs, the client **must** close the response body when finished with it.

Integralist:
Fixes #362

### DIFF
--- a/fastly/acl.go
+++ b/fastly/acl.go
@@ -52,6 +52,7 @@ func (c *Client) ListACLs(i *ListACLsInput) ([]*ACL, error) {
 	if err != nil {
 		return nil, err
 	}
+	defer resp.Body.Close()
 
 	var as []*ACL
 	if err := decodeBodyMap(resp.Body, &as); err != nil {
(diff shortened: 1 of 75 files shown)


doramatadora (approved):
Surprisingly easy to review — 1 change in many places :)

joeshaw:
No objection here, but `decodeBodyMap` already closes the body and it looks like almost all (if not all) requests call that: https://github.com/fastly/go-fastly/blob/fedbcfc1903eca480f41e62a5f47619fda1108a3/fastly/client.go#L471

That makes me wonder if the leak was somewhere else, or if there were particular calls that weren't closing the body.

And it means that the `defer body.Close()` can probably be dropped from `decodeBodyMap` and it can take an `io.Reader` instead of an `io.ReadCloser`.

Integralist:
Damn, I didn't realise it was in the `decodeBodyMap`. My bad. I'll open another PR to fix this up and remove from decodeBodyMap as you've suggested.

PR #365, opened 2022-09-08
Avoid duplicate Close() call on http.Response.Body

Integralist:
When fixing https://github.com/fastly/go-fastly/issues/362 I neglected to notice that the `decodeBodyMap` function was already calling `.Close()` on the provided `http.Response.Body`.

### DIFF
--- a/fastly/client.go
+++ b/fastly/client.go
@@ -467,9 +467,7 @@ func checkResp(resp *http.Response, err error) (*http.Response, error) {
 
 // decodeBodyMap is used to decode an HTTP response body into a mapstructure struct.
 // It closes `body`.
-func decodeBodyMap(body io.ReadCloser, out interface{}) error {
-	defer body.Close()
-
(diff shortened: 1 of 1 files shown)


fastly/client.go
```diff
@@ -467,9 +467,7 @@ func checkResp(resp *http.Response, err error) (*http.Response, error) {
 
 // decodeBodyMap is used to decode an HTTP response body into a mapstructure struct.
 // It closes `body`.
```
joeshaw, line 469: Comment needs updating.

joeshaw (changes requested):

joeshaw:
Sorry, got ahead of myself approving before I noticed the comment needed updating.

fastly/client.go
Integralist: 🤦🏻

joeshaw (approved):
👍 for real now

Issue #371, opened 2022-10-06
UpdateBackend() can't clear OverrideHost and other fields?


dkegel-fastly:
The omitempty in UpdateBackendInput appears to prevent callers from clearing OverrideHost and other fields.

bdournaee:
This is affecting a customer using the signal sciences edge deployment. Any help appreciated in addressing it.

dkegel-fastly:
Looks like I was confused, and the real problem I was trying to solve lies elsewhere.

PR #372, opened 2022-10-06
backends: allow clearing at least some fields on update

dkegel-fastly:
Partial fix for https://github.com/fastly/go-fastly/issues/371 ?

### DIFF
--- a/fastly/backend.go
+++ b/fastly/backend.go
@@ -203,11 +203,14 @@ type UpdateBackendInput struct {
 	// Name is the name of the backend to update.
 	Name string
 
+	// Which fields should be omitempty?  NewName certainly,
+	// but not e.g. OverrideHost and SSLSNIHostname, which users may want to clear.
+	// TODO: remove omitempty from more fields
 	NewName             *string      `url:"name,omitempty"`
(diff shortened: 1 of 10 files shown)


dgryski:
I'm concerned about this causing subtle breakage for our users.

dkegel-fastly:
Me, too.

Integralist:
> I'm concerned about this causing subtle breakage for our users.

We can just do a major version update then, no? 

[This typo](https://github.com/fastly/go-fastly/blob/41260b8314dd09b4a9265260a461f94c5c5850b5/fastly/service_authorization.go#L30) has been bugging me for a bit so I could fix that and maybe some other breaking changes as part of the next major release.

dgryski:
If we have a bunch of breaking changes lined up, that's not a bad plan.  We could write some tooling (semgrep/ruleguard) to detect/fix the common ones.

dkegel-fastly:
One breaking change might be "generate this sdk with https://github.com/fastly/api-clients-generator/ "

Integralist:
Closing in favour of https://github.com/fastly/go-fastly/pull/375

PR #373, opened 2022-10-10
Support Health Check Headers

Integralist:
Example API call via `curl`:

```shell
curl -i -X POST "https://api.fastly.com/service/123/version/2/healthcheck" \
  -H "Fastly-Key: $FASTLY_API_KEY" \
  -H "Content-Type: application/x-www-form-urlencoded" \
  -H "Accept: application/json" \
  -d 'name=test-healthcheck4&host=example.com&path=%2Ftest.txt&headers=["Foo: Bar","Beep: Boop"]'
```

### DIFF
--- a/fastly/client.go
+++ b/fastly/client.go
@@ -317,6 +317,32 @@ func (c *Client) Request(verb, p string, ro *RequestOptions) (*http.Response, er
 	return resp, nil
 }
 
+// parseHealthCheckHeaders returns the serialised body with the custom health
+// check headers appended.
+//
+// NOTE: The Google query library we use for parsing and encoding the provided
+// struct values doesn't support the format `headers=["Foo: Bar"]` and so we
+// have to manually construct this format.
+func parseHealthCheckHeaders(s string) string {
+	headers := []string{}
+	result := []string{}
+	segs := strings.Split(s, "&")
+	for _, s := range segs {
+		if strings.HasPrefix(strings.ToLower(s), "headers=") {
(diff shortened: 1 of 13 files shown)


fastly/client.go
```diff
@@ -317,6 +317,32 @@ func (c *Client) Request(verb, p string, ro *RequestOptions) (*http.Response, er
 	return resp, nil
 }
 
+// parseHealthCheckHeaders returns the serialised body with the custom health
+// check headers appended.
+//
+// NOTE: The Google query library we use for parsing and encoding the provided
+// struct values doesn't support the format `headers=["Foo: Bar"]` and so we
+// have to manually construct this format.
```
Integralist, line 325: ℹ️ We had to use the Google library for serialising data because it was the only library that supported a separate form of array syntax our APIs depend on. But unfortunately, the new health check API uses a syntax that even the Google library doesn't support.

fastly/client.go
```diff
@@ -333,7 +359,11 @@ func (c *Client) RequestForm(verb, p string, i interface{}, ro *RequestOptions)
 	if err != nil {
 		return nil, err
 	}
+
 	body := v.Encode()
+	if ro.HealthCheckHeaders {
```
Integralist, line 364: ℹ️ To reduce the blast radius of this type of manual serialisation I've locked it down to just the CREATE and UPDATE API endpoints for health checks.

triblondon (approved):

PR #375, opened 2022-10-24
Backend resource tweaks

Integralist:
Issue https://github.com/fastly/go-fastly/issues/371 suggests setting `override_host` to an empty value is not updating the field. The existing test suite demonstrates that `override_host` can be overridden, and additional testing has shown that the field can be updated with a zero value if required.

This PR makes some amendments to the test suite, and also implements a breaking-change by modifying the input type of the port field.

## NOTES

To summarise the pre-existing implementation details...

If a struct field isn't populated, and is marshalled to JSON, then the field's zero value will be used (e.g. type string zero value == "", type int zero value == 0, etc).

You can use `omitempty` to prevent the field from being marshalled, but then you won't know if the zero value was intentional or not (e.g. a user might _want_ to set an type int field to zero or a type string field to an empty string).

To avoid that situation you need to have the field be set to a _pointer_ of the type. This is because the zero value for a pointer is `nil`. This means if the field is `nil` then the field was never set and won't be marshalled, but if the pointer points to a value, then we know it was set to the zero value intentionally.

> Reference: https://willnorris.com/2014/05/go-rest-apis-and-pointers/

### DIFF
--- a/fastly/backend.go
+++ b/fastly/backend.go
@@ -221,7 +221,7 @@ type UpdateBackendInput struct {
 	SSLCACert           *string      `url:"ssl_ca_cert,omitempty"`
 	SSLCertHostname     *string      `url:"ssl_cert_hostname,omitempty"`
 	SSLCheckCert        *Compatibool `url:"ssl_check_cert,omitempty"`
-	SSLCiphers          string       `url:"ssl_ciphers,omitempty"`
+	SSLCiphers          *string      `url:"ssl_ciphers,omitempty"`
 	SSLClientCert       *string      `url:"ssl_client_cert,omitempty"`
 	SSLClientKey        *string      `url:"ssl_client_key,omitempty"`
 	SSLSNIHostname      *string      `url:"ssl_sni_hostname,omitempty"`
--- a/fastly/backend.go
+++ b/fastly/backend.go
@@ -12,36 +12,36 @@ type Backend struct {
 	ServiceID      string `mapstructure:"service_id"`
 	ServiceVersion int    `mapstructure:"version"`
 
-	Name                string     `mapstructure:"name"`
-	Comment             string     `mapstructure:"comment"`
 	Address             string     `mapstructure:"address"`
-	Port                uint       `mapstructure:"port"`
-	OverrideHost        string     `mapstructure:"override_host"`
+	AutoLoadbalance     bool       `mapstructure:"auto_loadbalance"`
+	BetweenBytesTimeout uint       `mapstructure:"between_bytes_timeout"`
+	Comment             string     `mapstructure:"comment"`
 	ConnectTimeout      uint       `mapstructure:"connect_timeout"`
-	MaxConn             uint       `mapstructure:"max_conn"`
+	CreatedAt           *time.Time `mapstructure:"created_at"`
+	DeletedAt           *time.Time `mapstructure:"deleted_at"`
 	ErrorThreshold      uint       `mapstructure:"error_threshold"`
 	FirstByteTimeout    uint       `mapstructure:"first_byte_timeout"`
-	BetweenBytesTimeout uint       `mapstructure:"between_bytes_timeout"`
-	AutoLoadbalance     bool       `mapstructure:"auto_loadbalance"`
-	Weight              uint       `mapstructure:"weight"`
-	RequestCondition    string     `mapstructure:"request_condition"`
 	HealthCheck         string     `mapstructure:"healthcheck"`
 	Hostname            string     `mapstructure:"hostname"`
-	Shield              string     `mapstructure:"shield"`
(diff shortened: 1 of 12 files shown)


fastly/backend.go
```diff
@@ -96,36 +96,36 @@ type CreateBackendInput struct {
...
-	SSLSNIHostname  string      `url:"ssl_sni_hostname,omitempty"`
-	MinTLSVersion   string      `url:"min_tls_version,omitempty"`
-	MaxTLSVersion   string      `url:"max_tls_version,omitempty"`
-	SSLCiphers      string      `url:"ssl_ciphers,omitempty"`
+	MaxConn             *uint       `url:"max_conn,omitempty"`
+	MaxTLSVersion       string      `url:"max_tls_version,omitempty"`
+	MinTLSVersion       string      `url:"min_tls_version,omitempty"`
+	Name                string      `url:"name,omitempty"`
+	OverrideHost        string      `url:"override_host,omitempty"`
+	Port                uint        `url:"port,omitempty"`
```
Integralist: ℹ️ I changed the `Port` field from a pointer to primitive as I don't believe anyone would want to set the port number to zero intentionally and so having the zero value omitted in this case should be fine.

fastly/backend.go
```diff
@@ -203,32 +203,31 @@ type UpdateBackendInput struct {
...
 	SSLClientKey        *string      `url:"ssl_client_key,omitempty"`
-	SSLHostname         *string      `url:"ssl_hostname,omitempty"`
-	SSLCertHostname     *string      `url:"ssl_cert_hostname,omitempty"`
 	SSLSNIHostname      *string      `url:"ssl_sni_hostname,omitempty"`
-	MinTLSVersion       *string      `url:"min_tls_version,omitempty"`
-	MaxTLSVersion       *string      `url:"max_tls_version,omitempty"`
-	SSLCiphers          string       `url:"ssl_ciphers,omitempty"`
+	Shield              *string      `url:"shield,omitempty"`
+	UseSSL              *Compatibool `url:"use_ssl,omitempty"`
+	Weight              *uint        `url:"weight,omitempty"`
```
Integralist, line 226: ℹ️ The changes to this struct are just sorting the fields in alphabetical order.

fastly/backend_test.go
```diff
@@ -21,7 +21,6 @@ func TestClient_Backends(t *testing.T) {
 			ServiceVersion: tv.Number,
 			Name:           "test-backend",
 			Address:        "integ-test.go-fastly.com",
-			Port:           Uint(1234),
```
Integralist, line 24: ℹ️ I removed the setting of a port on creation so we can validate the API is setting a default of `80`.

fastly/backend.go
```diff
@@ -12,36 +12,36 @@ type Backend struct {
 	ServiceID      string `mapstructure:"service_id"`
 	ServiceVersion int    `mapstructure:"version"`
 
-	Name                string     `mapstructure:"name"`
```
Integralist, line 15: ℹ️ I sorted the fields

fastly/backend_test.go
```diff
@@ -63,6 +62,9 @@ func TestClient_Backends(t *testing.T) {
 	if b.OverrideHost != "origin.example.com" {
 		t.Errorf("bad override_host: %q", b.OverrideHost)
 	}
+	if b.SSLCheckCert != true {
```
Integralist, line 65: ℹ️ I also confirm the API defaults to setting the `ssl_check_cert` field to `true`.

Integralist (commented):
Just some comments to help the PR reviewer...

fastly/backend_test.go
```diff
@@ -21,10 +21,10 @@ func TestClient_Backends(t *testing.T) {
 			ServiceVersion: tv.Number,
 			Name:           "test-backend",
 			Address:        "integ-test.go-fastly.com",
-			Port:           Uint(1234),
 			ConnectTimeout: Uint(1500),
 			OverrideHost:   "origin.example.com",
 			SSLCiphers:     "DHE-RSA-AES256-SHA:DHE-RSA-CAMELLIA256-SHA:AES256-GCM-SHA384",
+			SSLSNIHostname: "ssl-hostname.com",
```
Integralist, line 27: ℹ️ I've added a test assertion for `SSLSNIHostname` as that was also mentioned as being problematic on the Fastly customer support ticket.

fastly/backend.go
dgryski: Note changing the field order will break users who don't use https://github.com/uber-go/guide/blob/master/style.md#use-field-names-to-initialize-structs .  (`go vet` catches this)

Integralist: I indeed to make other breaking changes BTW before cutting the next major release.

fgsch (approved):
Untested by code LGTM 👍

PR #381, opened 2022-11-03
Remove read-only fields from Create/Update structs

Integralist:


Integralist:
Discovered although the docs say READ-ONLY the S3 tests indicate that the field can be set!

PR #385, opened 2022-11-14
fastly: fix invalid package uploading test

glerchundi:
It seems like the expected response from Fastly API slightly changed since it was added the test for invalid packages. Now, instead of returning package info with zeroed properties, it properly returns a specific HTTP error (bad request).

This PR changes the test to be compliant with the new behaviour.

Signed-off-by: Gorka Lerchundi Osa <glertxundi@gmail.com>

### DIFF
--- a/fastly/fixtures/package/get.yaml
+++ b/fastly/fixtures/package/get.yaml
@@ -6,21 +6,21 @@ interactions:
     form: {}
     headers:
       User-Agent:
-      - FastlyGo/2.1.0 (+github.com/fastly/go-fastly; go1.15.4)
-    url: https://api.fastly.com/service/7i6HN3TK9wS159v2gPAZ8A/version/2/package
+      - FastlyGo/7.0.0 (+github.com/fastly/go-fastly; go1.18.3)
+    url: https://api.fastly.com/service/I1VEQVKrgJFQszKVNY3FG7/version/2/package
     method: GET
   response:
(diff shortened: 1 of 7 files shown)


fastly/fixtures/package/get.yaml
```diff
@@ -6,21 +6,21 @@ interactions:
     form: {}
     headers:
       User-Agent:
-      - FastlyGo/2.1.0 (+github.com/fastly/go-fastly; go1.15.4)
-    url: https://api.fastly.com/service/7i6HN3TK9wS159v2gPAZ8A/version/2/package
+      - FastlyGo/7.0.0 (+github.com/fastly/go-fastly; go1.18.3)
+    url: https://api.fastly.com/service/pChqKASI98hryShq2mlDf6/version/2/package
```
Integralist, line 10: If you could run `make fix-fixtures FASTLY_TEST_SERVICE_ID="pChqKASI98hryShq2mlDf6"` and that should replace any reference to your Service ID with the go-fastly generic one.

Integralist (changes requested):
One small change and then I'll approve and merge. Thanks for discovering this!

fastly/fixtures/package/get.yaml
glerchundi: I was pretty sure that I executed it 🤦. Probably I messed up between different test runs. Anyway, fixed and ready to review!

PR #386, opened 2022-11-15
Improve error response handling

awilliams-fastly:
Instead of `panic`'ing when an error response cannot be decoded from JSON, return a generic error which includes the read portion of the response body.

Followup to @dgryski's comment: https://github.com/fastly/go-fastly/pull/367#discussion_r980318242

### DIFF
--- a/fastly/errors.go
+++ b/fastly/errors.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"strconv"
 
@@ -327,11 +328,23 @@ func NewHTTPError(resp *http.Response) *HTTPError {
 		return &e
 	}
 
+	// Save a copy of the body as it's read/decoded.
+	// If decoding fails, it can then be used (via addDecodeErr)
+	// to create a generic error containing the body's read contents.
+	var bodyCp bytes.Buffer
+	body := io.TeeReader(resp.Body, &bodyCp)
+	addDecodeErr := func() {
+		e.Errors = append(e.Errors, &ErrorObject{
+			Title:  "Undefined error",
+			Detail: bodyCp.String(),
+		})
+	}
+
 	switch resp.Header.Get("Content-Type") {
 	case jsonapi.MediaType:
-		// If this is a jsonapi response, decode it accordingly
-		if err := decodeBodyMap(resp.Body, &e); err != nil {
-			panic(err)
+		// If this is a jsonapi response, decode it accordingly.
+		if err := decodeBodyMap(body, &e); err != nil {
+			addDecodeErr()
 		}
 
 	case "application/problem+json":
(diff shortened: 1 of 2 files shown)


fastly/errors.go
```diff
@@ -327,11 +328,23 @@ func NewHTTPError(resp *http.Response) *HTTPError {
 		return &e
 	}
 
+	// Save a copy of the body as it's read/decoded.
+	// If decoding fails, it can then be used (via addDecodeErr)
+	// to create a generic error containing the body's read contents.
+	var bodyCp bytes.Buffer
```
dgryski, line 334: I'm not sure we need to save a copy of the entire body in case of error.  Any decoding error "should" have enough context on its own.

dgryski (changes requested):
I'd like more details on what the error message improvement is from keeping the entire body in memory.

fastly/errors.go
awilliams-fastly: This assumes that the response body contains useful text (that's either not JSON or is malformed JSON) for end users. Such as in the contrived [test case](https://github.com/fastly/go-fastly/pull/386/files#diff-3e88c7dd341d9c3ca93ea88df31f1d05b321d559d0b21eaec23c6c7047fa4e2eR130-R134):

```
404 - Not Found:
    Title:  Undefined error
    Detail: THIS IS NOT JSON
```

I'm not sure of the utility of returning the decode error. Taking the same test example, the error message would become:

```
404 - Not Found:

    Title:  Undefined error
    Detail: invalid character 'T' looking for beginning of value
```

fastly/errors.go
```diff
@@ -327,11 +328,23 @@ func NewHTTPError(resp *http.Response) *HTTPError {
...
+	body := io.TeeReader(resp.Body, &bodyCp)
+	addDecodeErr := func() {
```
dgryski, line 336: In all the places this is called, we actually have a non-nil `err` variable.  I wonder if we could include it in the error object somehow (instead of a generic `Undefined error`).

dgryski (changes requested):
One more question for the error text, but once that's resolved we're good to go.

fastly/errors.go
dgryski: Thanks for the justification.

dgryski (approved):
LGTM

fastly/errors.go
awilliams-fastly: This is possible, and so I wired it up locally and tested.

As an example, imagine that `api.fastly.com` returns an `Fastly error: unknown domain.` plaintext error response body. Users would see something like the following:

```
502 - Bad Gateway

    Title:  Undefined error
    Detail: Fastly error: unknown domain.
    Meta:   map[json_decode_error: invalid character 'F' looking for beginning of value]
```

I'm afraid that the decode error could be conflated with the actual error, leading to confusion.

PR #392, opened 2023-02-03
add support for Secret Store client keys

joeshaw:
This lets clients create a client key for local encryption of secrets
before uploading to the Fastly API.

This bumps the minimum required Go version to ~1.17~ 1.18 and introduces
`golang.org/x/crypto` as a dependency.


joeshaw:
@Integralist @awilliams-fastly I don't have write access to this repo so I pushed to a fork.  I also can't set reviewers so I'm tagging you in!

joeshaw:
Bleh, the linter wants Go 1.19.  I'm not entirely sure why the version on `honnef.co/go/tools` got bumped.  I suspect it has to do with bringing in the crypto package.

Let me know what you want me to do.  I can either try to find an older version of the crypto package that works with the older tools package, or we can bump up and require Go 1.19.

Even if we go with the older version of the crypto package, we will need to bump the requirement to Go 1.17 because the crypto API requires casting a slice to a pointer of a fixed-size array.

Integralist:
👋🏻 @joeshaw you've been given WRITE access to this repo now.

It's been suggested that the API client should be using two versions back from the latest release, so that would suggest 1.18 is fine to bump to here if you want to (or if that helps with finding an older crypto package).

fastly/secret_store.go
```diff
@@ -379,3 +384,77 @@ func (c *Client) DeleteSecret(i *DeleteSecretInput) error {
...
+// using the public signing key before it is used.  A ValidateSignature
+// method is provided for this purpose.
+type ClientKey struct {
+	PublicKey []byte    `json:"public_key"`
+	Signature []byte    `json:"signature"`
+	ExpiresAt time.Time `json:"expires_at"`
+}
+
+// ValidateSignature returns true if the client key's signature for the
+// provided public signing key.
```
Integralist, line 404: This sentence feels incomplete.

I had a look at the `ed25519` package to get a sense of what its `Verify` does and came up with the following:

```suggestion
// ValidateSignature reports if signingKey was used to 
// generate the signature. It will panic if len(publicKey) 
// is not ed25519.PublicKeySize.
// https://pkg.go.dev/crypto/ed25519@go1.19.5#PublicKeySize
```

fastly/secret_store_test.go
```diff
@@ -171,6 +176,79 @@ func TestClient_CreateSecret(t *testing.T) {
...
+
+	if got := ck.PublicKey; len(got) == 0 {
+		t.Errorf("PublicKey: got empty")
+	}
+
+	if got := ck.Signature; len(got) == 0 {
+		t.Errorf("Signature: got empty")
+	}
+
+	if got, now := ck.ExpiresAt, time.Now(); got.Before(now) {
```
Integralist, line 204: Will there be issues using `time.Now()` considering we're producing a static fixture?

joeshaw: I was looking in the codebase for other similar checks and didn't find any.  It doesn't look like the date is encoded in the test fixtures at all (other than a `Date` response header).  Would it be ok to hardcode a date in the test (resetting it to now if `vcrDisabled` is set)?  Or should I loosen this check to just seeing if the value is present?  Or drop the check altogether?

joeshaw: Actually, never mind.  I think I can use the server's Date response header to calibrate things.

joeshaw: Er, no, ignore that last comment.  I don't have access to the response headers at the time I'm checking.

Integralist: I would maybe just avoid this check for now unless we have a good way to handle it for vcr type fixtures.

joeshaw:
Thanks for the review!  I've rebased on main, fixed up the module versioning, and made Go 1.18 the minimum required version.

The CI seems to be failing with the `make check-mod` target, but I can't reproduce the failure locally.  `make check-mod` and `go mod tidy` are clean for me (trying 1.18.10 and 1.19.5)

joeshaw:
Ok, I'm able to reproduce the CI failure if I clear out my module cache with `go clean -modcache`.  I think the difference in behavior is due to this (from `go help mod download`):

> With no arguments, download applies to the modules needed to build and test
the packages in the main module: the modules explicitly required by the main
module if it is at 'go 1.17' or higher, or all transitively-required modules
if at 'go 1.16' or lower.

(Previously the `go.mod` file was `go 1.16`.)

joeshaw:
Ok, latest commit fixes the CI error.  `go mod tidy -v` downloading modules is fine.  It's only when a module should be added or removed (not downloaded) that it should error out.  So we now filter out downloading messages from the output, and we no longer need to `go mod download` as a separate step.

joeshaw:
Finally CI passes! 😅 staticcheck had to be upgraded to 2022.1.3 support generics (Go 1.18) but couldn't be updated to the latest release (2023.1) because it requires Go 1.19.

fastly/secret_store.go
```diff
@@ -379,3 +388,91 @@ func (c *Client) DeleteSecret(i *DeleteSecretInput) error {
...
+// ValidateSignature reports if the signingKey was used to generate the
+// client key's signature.  It must be a valid Ed25519 public key, and
+// it will panic if len(signingKey) is not ed25519.PublicKeySize.
+// https://pkg.go.dev/crypto/ed25519#PublicKeySize
+func (ck *ClientKey) ValidateSignature(signingKey ed25519.PublicKey) bool {
+	return ed25519.Verify(signingKey, ck.PublicKey, ck.Signature)
+}
+
+// Encrypt uses the client key to encrypt the provided plaintext
+// using a libsonium-compatible sealed box.
```
awilliams-fastly, line 416:
```suggestion
// using a libsodium-compatible sealed box.
```

awilliams-fastly (approved):

fastly/secret_store.go
```diff
@@ -379,3 +388,91 @@ func (c *Client) DeleteSecret(i *DeleteSecretInput) error {
...
+func (ck *ClientKey) ValidateSignature(signingKey ed25519.PublicKey) bool {
+	return ed25519.Verify(signingKey, ck.PublicKey, ck.Signature)
+}
+
+// Encrypt uses the client key to encrypt the provided plaintext
+// using a libsodium-compatible sealed box.
+// https://pkg.go.dev/golang.org/x/crypto/nacl/box#SealAnonymous
+// https://libsodium.gitbook.io/doc/public-key_cryptography/sealed_boxes
+func (ck *ClientKey) Encrypt(plaintext []byte) ([]byte, error) {
+	if len(ck.PublicKey) != 32 {
```
Integralist, line 420: Should we use `ed25519.PublicKeySize` instead of the 'magic number'?
```suggestion
	if len(ck.PublicKey) != ed25519.PublicKeySize {
```

joeshaw: No, because these are actually not the same thing.  The _signing key_ is an Ed25519 public key, and uses the APIs from https://pkg.go.dev/crypto/ed25519.  The _client key_ is an X25519 public key, and uses the APIs from https://pkg.go.dev/golang.org/x/crypto/nacl/box.  They're derived from similar things but aren't quite the same.  [This Stack Overflow answer](https://crypto.stackexchange.com/a/68129) goes into some details on the differences.

Both are 32 bytes long so it would work, but it doesn't feel quite right to use the `ed25519` values here.  Unfortunately, the `nacl/box` package hardcodes 32-byte arrays throughout its API and doesn't define a constant for this.

Integralist (approved):

PR #395, opened 2023-02-08
feat(http3): implement HTTP3 API endpoints

Integralist:


fastly/http3.go
```diff
@@ -0,0 +1,115 @@
...
+
+// HTTP3 represents a response from the Fastly API.
+type HTTP3 struct {
+	CreatedAt       *time.Time `mapstructure:"created_at"`
+	DeletedAt       *time.Time `mapstructure:"deleted_at"`
+	FeatureRevision int        `mapstructure:"feature_revision"`
+	ServiceID       string     `mapstructure:"service_id"`
+	ServiceVersion  int        `mapstructure:"version"`
+	UpdatedAt       *time.Time `mapstructure:"updated_at"`
+}
```
awilliams-fastly, line 16: It may be useful to add corresponding `json` field tags so that `json.Marshal(HTTP3)` works as expected.

fastly/http3.go
```diff
@@ -0,0 +1,115 @@
...
+	}
+
+	path := fmt.Sprintf("/service/%s/version/%d/http3", i.ServiceID, i.ServiceVersion)
+	resp, err := c.Get(path, nil)
+	if err != nil {
+		return nil, err
+	}
+	defer resp.Body.Close()
+
+	var h *HTTP3
```
awilliams-fastly, line 42: May be worth adding a comment as to why it's `*HTTP3` and not `HTTP3`

awilliams-fastly (approved):

fastly/http3.go
Integralist: Thanks @awilliams-fastly yup I saw you doing this recently and that sounds like a good general practice 👍🏻

PR #397, opened 2023-02-10
secret-store: rename ValidateSignature to VerifySignature

joeshaw:
A more technically correct term.

### DIFF
--- a/fastly/secret_store.go
+++ b/fastly/secret_store.go
@@ -404,11 +404,11 @@ type ClientKey struct {
 	ExpiresAt time.Time `json:"expires_at"`
 }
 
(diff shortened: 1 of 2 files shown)


Integralist (approved):
LGTM but this will be a change in the exposed interface and so must be treated as a major version bump (i.e. to release this change we'd bump to `8.0.0`). So if there are any other 'breaking' changes you think could do with made, then now would be the time to get those in.

joeshaw:
In that case, let's sit on this for now.  It's not a big deal (although I also guarantee nobody is using the old name so it's largely a technicality) so it's not worth making a major version bump for this alone.

awilliams-fastly (approved):

PR #398, opened 2023-02-16
feat: Add Config Store support

awilliams-fastly:
- https://developer.fastly.com/reference/api/services/resources/config-store/
- https://developer.fastly.com/reference/api/services/resources/config-store-item/

Related `cli` change: https://github.com/fastly/cli/pull/829

### DIFF
--- a/fastly/config_store.go
+++ b/fastly/config_store.go
@@ -1,0 +1,250 @@
+package fastly
+
+import (
+	"encoding/json"
+	"fmt"
+	"sort"
+	"time"
+)
+
+// Config Store.
+// A container that lets you store data in key-value pairs.
+// https://developer.fastly.com/reference/api/services/resources/config-store/
+// https://developer.fastly.com/reference/api/services/resources/config-store-item/
+
+// ConfigStore represents a config store response from the Fastly API.
+type ConfigStore struct {
+	Name      string     `json:"name"`
+	ID        string     `json:"id"`
+	CreatedAt *time.Time `json:"created_at"`
+	UpdatedAt *time.Time `json:"updated_at"`
+	DeletedAt *time.Time `json:"deleted_at"`
+}
+
+// ConfigStoreMetadata represents a config store metadata response from the Fastly API.
+type ConfigStoreMetadata struct {
+	// ItemCount is the number of items in a store.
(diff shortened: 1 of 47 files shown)


fastly/config_store_item_test.go
```diff
@@ -0,0 +1,267 @@
...
+			StoreID: cs.ID,
+			Key:     t.Name(),
+			Value:   value,
+		})
+	})
+	if err != nil {
+		t.Fatalf("error creating config store item: %v", err)
+	}
+
+	if got, want := item.StoreID, cs.ID; got != want {
```
Integralist, line 30: Just for my own understanding, is it more common to use `have, want` or `got, want` I feel like the former is better due to both words having the same length although `got` is shorter (and that's usually preferred). 

This is more 'bike shedding' than anything so not overly important but from a 'style' perspective I was interested in what we should use for consistency.

Integralist (commented):
LGTM. Just a couple of comment.

fastly/config_store_item_test.go
awilliams-fastly: I don't think this is a particularly strong convention, but `got/want` is mentioned in the [TestComments](https://github.com/golang/go/wiki/TestComments#got-before-want) Go wiki and is what I see most often.

fastly/config_store.go
```diff
@@ -0,0 +1,250 @@
...
+
+// CreateConfigStoreInput is the input to CreateConfigStore.
+type CreateConfigStoreInput struct {
+	// Name is the name of the store to create (required).
+	Name string `url:"name"`
+}
+
+// CreateConfigStore creates a new Fastly config store.
+func (c *Client) CreateConfigStore(i *CreateConfigStoreInput) (*ConfigStore, error) {
+	if i.Name == "" {
```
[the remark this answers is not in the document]
awilliams-fastly: Thanks for the context!

Updated in e6d490a to only validate URL path arguments.

fastly/config_store_item_test.go
Integralist: Good enough for me 🙂 👍🏻

Integralist (approved):

PR #403, opened 2023-03-01
fix(client): add title to legacy error format

Integralist:
Some APIs (e.g. Product Enablement) return a data structure compatible with `application/problem+json` but not the actual Content-Type that would cause go-fastly to be able to parse the data. 

The following code example is the scenario I'm trying to reach. This because we have discovered a buggy workflow in the Terraform provider and to resolve it we need to be able to determine if the user has 'self-enablement' turned on for their customer account. Without the changes in this PR we are unable to determine that because the status code by itself (400) is too generic.

We can't access the full message from Terraform as it chops off the message...
Screenshot 2023-03-01 at 17 38 41

By exposing the title we can convert to HTTPError and check the string directly...
```go
package main

import (
	"fmt"
	"log"
	"strings"

	"github.com/fastly/go-fastly/v7/fastly"
)

func main() {
	client, err := fastly.NewClient("<REDACTED>")
	if err != nil {
		log.Fatal(err)
	}

	err = client.DisableProduct(&fastly.ProductEnablementInput{
		ProductID: fastly.ProductImageOptimizer,
		ServiceID: "<REDACTED>",
	})
	if err != nil {
		if he, ok := err.(*fastly.HTTPError); ok {
			for _, e := range he.Errors {
				if strings.Contains(e.Title, "is not entitled to disable product") {
					fmt.Println(e.Title)
					return
				}
			}
		}
		log.Fatal(err)
	}
}
```

### DIFF
--- a/fastly/errors.go
+++ b/fastly/errors.go
@@ -308,12 +308,13 @@ type HTTPError struct {
 
 // ErrorObject is a single error.
 type ErrorObject struct {
-	Code   string                  `mapstructure:"code"`
-	Detail string                  `mapstructure:"detail"`
-	ID     string                  `mapstructure:"id"`
-	Meta   *map[string]interface{} `mapstructure:"meta"`
-	Status string                  `mapstructure:"status"`
-	Title  string                  `mapstructure:"title"`
+	Code    string                  `mapstructure:"code"`
+	Detail  string                  `mapstructure:"detail"`
+	ID      string                  `mapstructure:"id"`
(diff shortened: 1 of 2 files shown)


Integralist:
Actually, I've just realised I can effectively still get to that data from the original error object...

```go
if strings.Contains(err.Error(), "is not entitled to disable product") { ... }
```

Integralist:
Going to close this PR in favour of fixing the API itself.

Issue #408, opened 2023-03-09
Lack of support for warnings in TLSAuthorizations struct


lukaszczerpak-cloudinary:
API can return warnings in `tls_authorization` object (ie. when DV cannot be completed):

```json
...
        "created_at": "2023-02-04T00:00:00.000Z",
        "state": "passing",
        "updated_at": "2023-02-05T00:00:00.000Z",
        "warnings": [
          {
            "type": "caa",
            "instructions": "Found conflicting CAA record(s). Expected globalsign.com"
          }
        ]
...
```

This information is not serialised to golang struct here:
https://github.com/fastly/go-fastly/blob/main/fastly/tls_subscription.go#L32

The status itself `passing` doesn't indicate that something is wrong, so based on purely information in the struct it's impossible to know there is a problem anywhere.

The `warnings` array should be added to the struct and ideally new status crated to cover this situation, ie. `warning`.

PR #409, opened 2023-03-14
fix(tls-subscription): serialise warnings field

Integralist:
Fixes: https://github.com/fastly/go-fastly/issues/408

We should expose `warnings` field, as a "passing" state doesn't always indicate a successful operation.

> **NOTE:** I've internally raised a question related to why the API returns a `200 OK` when it seems a warning (as reported in the customer issue, link above) suggests that in fact this would cause the TLS subscription operation to fail.


awilliams-fastly (approved):
👍

PR #422, opened 2023-04-11
breaking(object-store): rename ObjectStore to KVStore

dennismartensson:
We are renaming object store to kv store so doing the needed changes here.

This needs an internal API change and when that is out it should start working on a test rerun.


fastly/resource.go
```diff
@@ -81,12 +81,12 @@ func (c *Client) ListResources(i *ListResourcesInput) ([]*Resource, error) {
 
 // CreateResourceInput is used as input to the CreateResource function.
 type CreateResourceInput struct {
-	// Name is the name of the resource being linked to (e.g. an object store).
+	// Name is the name of the resource being linked to (e.g. an kv store).
```
Integralist, line 84: "an" is used before a word starting with a vowel, so with this change we should use "a" instead...

```suggestion
	// Name is the name of the resource being linked to (e.g. a kv store).
```

fastly/resource.go
```diff
@@ -157,7 +157,7 @@ func (c *Client) GetResource(i *GetResourceInput) (*Resource, error) {
 
 // UpdateResourceInput is used as input to the UpdateResource function.
 type UpdateResourceInput struct {
-	// Name is the name of the resource being linked to (e.g. an object store).
+	// Name is the name of the resource being linked to (e.g. an kv store).
```
Integralist, line 160: "an" is used before a word starting with a vowel, so with this change we should use "a" instead...
```suggestion
	// Name is the name of the resource being linked to (e.g. a kv store).
```

Integralist (changes requested):
Looks like we've made a change that shouldn't have been made...

```
fastly/response_object.go → fastly/response_kv.go
```

That file (and its associated test) is not related to the Object Store but to a VCL 'response object': https://developer.fastly.com/reference/api/vcl-services/response-object/

fastly/kv_store.go
```diff
@@ -0,0 +1,407 @@
...
+
+import (
+	"io"
+	"net/http"
+	"strconv"
+	"strings"
+	"time"
+)
+
+// https://developer.fastly.com/reference/api/kv-store/
```
Integralist, line 11:
```suggestion
// https://developer.fastly.com/reference/api/services/resources/kv-store
```

Integralist (approved):
LGTM.

PR #440, opened 2023-05-30
mTLS

Integralist:


fastly/tls_mutual_authentication.go
```diff
@@ -0,0 +1,212 @@
...
+	Enforced    bool             `jsonapi:"attr,enforced"`
+	ID          string           `jsonapi:"primary,mutual_authentication"`
+	Name        string           `jsonapi:"attr,name"`
+	UpdatedAt   *time.Time       `jsonapi:"attr,updated_at,iso8601"`
+}
+
+// ListTLSMutualAuthenticationsInput is used as input to the Client.ListTLSMutualAuthentication function.
+type ListTLSMutualAuthenticationsInput struct {
+	// Include is a comma-separated list of related objects to include (optional).
+	Include string
```
joeshaw, line 25: Any particular reason why this is `string` and not `[]string`?

fastly/tls_mutual_authentication.go
```diff
@@ -0,0 +1,212 @@
...
+				result[key] = value.(string)
+			}
+		case "int":
+			if value != 0 {
+				result[key] = strconv.Itoa(value.(int))
+			}
+		}
+	}
+
+	return result
```
joeshaw, line 54: Since there's only 3 things here, it's fewer lines of code to check them individually and avoid use of `reflect`:

```go
result := map[string]string{}
if i.Include != "" {
	result["include"] = i.Include
}
if i.PageSize > 0 {
	result["page[size]"] = strconv.Itoa(i.PageSize)
}
if i.PageNumber > 0 {
	result["page[number]"] = strconv.Itoa(i.PageNumber)
}
return result
```

fastly/tls_mutual_authentication.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	var ro *RequestOptions
+	if i.Include != "" {
+		ro = &RequestOptions{
+			Params: map[string]string{
+				"include": i.Include,
+			},
+		}
+	}
```
joeshaw, line 110: Should we set `Accept` here as in the list endpoint?

fastly/tls_mutual_authentication.go
```diff
@@ -0,0 +1,212 @@
...
+
+	return &o, nil
+}
+
+// CreateTLSMutualAuthenticationInput is used as input to the CreateTLSMutualAuthentication function.
+type CreateTLSMutualAuthenticationInput struct {
+	// CertBundle is one or more certificates. Enter each individual certificate blob on a new line. Must be PEM-formatted (required).
+	CertBundle string `jsonapi:"attr,cert_bundle"`
+	// ID should not be set (it's internally used to help marshal the JSONAPI request data).
+	ID string `jsonapi:"primary,mutual_authentication"`
```
joeshaw, line 132: I'm not familiar with JSON-API but is this a common convention?  Ideally we wouldn't be exposing internal fields like this in the API if we can avoid it.  One way to do this might be to make a private type inside `CreateTLSMutualAuthentication()` that embeds `CreateTLSMutualAuthenticationInput` and adds `ID`.

Something like:

```go
type CreateTLSMutualAuthenticationInput struct {
	CertBundle string `jsonapi:"attr,cert_bundle"`
	Name string `jsonapi:"attr,name,omitempty"`
}

func (c *Client) CreateTLSMutualAuthentication(i *CreateTLSMutualAuthenticationInput) (*TLSMutualAuthentication, error) {
	if i.CertBundle == "" {
		return nil, ErrMissingCertBundle
	}

	p := "/tls/mutual_authentications"

	in := struct{
		CreateTLSMutualAuthenticationInput
		ID `jsonapi:"primary,mutual_authentication"`
	}{
		CreateTLSMutualAuthenticationInput: i,
	}

	r, err := c.PostJSONAPI(p, in, nil)
	if err != nil {
		return nil, err
	}

	// ...
}
```

fastly/tls_mutual_authentication.go
```diff
@@ -0,0 +1,212 @@
...
+	// Name is a custom name for your mutual authentication (optional).
+	Name string `jsonapi:"attr,name,omitempty"`
+}
+
+// UpdateTLSMutualAuthentication updates the specified resource.
+//
+// By using this endpoint, the original certificate will cease to be used for future TLS handshakes.
+// Thus, only SAN entries that appear in the replacement certificate will become TLS enabled.
+// Any SAN entries that are missing in the replacement certificate will become disabled.
```
joeshaw, line 174: I don't think this text applies for this API.  If I am understanding this API, the cert bundle being uploaded is the set of CA certificates used to validate certificates being presented by TLS clients during the handshake.  If that's the case, SANs don't apply here.

fastly/tls_mutual_authentication.go
Integralist: > Ideally we wouldn't be exposing internal fields like this in the API if we can avoid it.

Agreed. 

I did this to keep it consistent with the rest of the code base which from what I can tell has always used this implementation.

I like the embed approach you've suggested as a way to internalise that ID 👍🏻 I'll give it a go locally and regenerate the fixtures and so if all works as expected it should just mean the dates in the regenerated fixtures change and nothing much else.

fastly/tls_mutual_authentication.go
Integralist: Hmm. I'm not completely sure but I as it's a JSON-API endpoint I don't think it would hurt to have the Accept header set.

fastly/tls_mutual_authentication.go
Integralist: In this particular instance, the API suggests there's only one permitted value (`tls_activations`). So although the documentation says comma-separated, that's not really true, but I have to presume the people who wrote the documentation were trying to be as 'future proof' as possible, so maybe we do the same and use a slice and join the strings for the user (just in case more permitted values are added in future).

fastly/tls_mutual_authentication.go
Integralist: @joeshaw This embedding approach doesn't appear to work btw...

```
tls_mutual_authentication_test.go:65: models should be a struct pointer or slice of struct pointers
```

Integralist: Just realised I wasn't passing a pointer `&in` so that moved the error onto...

```
    tls_mutual_authentication_test.go:65: 400 - Bad Request:

            Title:  Invalid value for cert_bundle
            Detail: can't be blank
```

Integralist: Spoke with Joe and he's OK with skipping this. If we could have made it work, then great but I tried 5 different ways to side-step the error stemming from the struct being embedded and ultimately it just wasn't compatible with how jsonapi marshalled the data.

Integralist:
Now that @joeshaw's other comments have been addressed I'm going to merge this PR later today.

joeshaw (approved):
👍

fastly/tls_mutual_authentication.go
```diff
@@ -0,0 +1,212 @@
...
+	if i.CertBundle == "" {
+		return nil, ErrMissingCertBundle
+	}
+
+	p := "/tls/mutual_authentications"
+
+	r, err := c.PostJSONAPI(p, i, nil)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
Integralist: Thanks! Will take a look 👍🏻

fastly/tls_mutual_authentication_test.go
```diff
@@ -0,0 +1,207 @@
...
+		t.Fatal(err)
+	}
+
+	cert, err := buildCertificate(privateKey, "example.com")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	// Create private key required to generate a custom certificate.
+	var pk *PrivateKey
```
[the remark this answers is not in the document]
Integralist: Thanks!

PR #453, opened 2023-06-29
fix: ensure Fastly-Key is stripped from the dump

Integralist:
To protect customers who use the `FASTLY_DEBUG_MODE` we should redact the Fastly-Key value.


joeshaw:
An alternative approach here might have been to clone the request with `req.Clone()` and then explicitly unsetting (or resetting to eg. `XXXXXX`) the `Fastly-Key` header.

PR #467, opened 2023-08-30
build: bump minimum go version to 1.19

Integralist:
The latest go version is 1.21 and I like to keep go-fastly two versions back from the latest.


dkegel-fastly (approved):
LGTM tho I wonder whether the staticcheck you're using needs updating...

PR #470, opened 2023-09-04
feat(auth): OAuth/PCKE callers

Integralist:
In a future world, callers (such as the Fastly CLI) will authenticate using OAuth (specifically the PKCE extension flow) and so the token that is provided will need to be used with the `Authentication` HTTP request header.

### DIFF
--- a/fastly/request.go
+++ b/fastly/request.go
@@ -1,6 +1,7 @@
 package fastly
 
 import (
+	"fmt"
 	"io"
 	"net/http"
 	"net/url"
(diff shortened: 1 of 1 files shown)


fastly/request.go
```diff
@@ -54,6 +55,7 @@ func (c *Client) RawRequest(verb, p string, ro *RequestOptions) (*http.Request,
 	// Set the API key.
 	if len(c.apiKey) > 0 {
 		request.Header.Set(APIKeyHeader, c.apiKey)
+		request.Header.Set("Authorization", fmt.Sprintf("Bearer %s", c.apiKey))
```
phamann, line 58: Should this be `c.apiKey`? That would set the value of the bearer to the API token and not the desired OAuth JWT `access_token`,

phamann (changes requested):

Integralist:
@phamann I've pinged you internally to explain my thought process. I'm going to close this PR as I think it's premature.

PR #472, opened 2023-09-28
fix(client): simplify key-redaction logic

Integralist:
This PR also ensures a response is dumped even when there is an error response.


joeshaw (approved):
👍 

(ref my comment in #453)

PR #481, opened 2023-11-07
feat(config store): adds ListConfigStoresInput 

stevendaniels:
* Adds ListConfigStoresInput, which allows for using the `name` param when retrieving a list of config stores.

### DIFF
--- a/fastly/config_store.go
+++ b/fastly/config_store.go
@@ -149,15 +149,28 @@ func (c *Client) GetConfigStoreMetadata(i *GetConfigStoreMetadataInput) (*Config
 	return csm, nil
 }
 
+// ListConfigStoreServicesInput is the input to ListConfigStoreServices.
+type ListConfigStoresInput struct {
+	// Name is the name of a config store (optional).
+	Name string
+}
+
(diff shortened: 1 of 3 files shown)


fastly/config_store.go
```diff
@@ -149,15 +149,28 @@ func (c *Client) GetConfigStoreMetadata(i *GetConfigStoreMetadataInput) (*Config
...
 
+// ListConfigStoreServicesInput is the input to ListConfigStoreServices.
+type ListConfigStoresInput struct {
+	// Name is the name of a config store (optional).
+	Name string
+}
+
 // ListConfigStores returns a list of config stores sorted by name.
-func (c *Client) ListConfigStores() ([]*ConfigStore, error) {
+func (c *Client) ListConfigStores(i *ListConfigStoresInput) ([]*ConfigStore, error) {
```
Integralist, line 159: This will result in a breaking interface change and a major release. Are we ok with doing that? If we're going to be pushing a breaking change we should check what else we can roll up into that release that is also breaking.

Integralist (approved):

PR #482, opened 2023-11-07
v8.7.0

stevendaniels:


CHANGELOG.md
```diff
@@ -1,5 +1,14 @@
 # Changelog
 
+## [v8.7.0](https://github.com/fastly/go-fastly/releases/tag/v8.7.0) (2023-11-07)
+
+[Full Changelog](https://github.com/fastly/go-fastly/compare/v8.6.4...v8.7.0)
+
+**Enhancements:**
+
+- feat(config stores): add `name` param for ListConfigStores` param [481](https://github.com/fastly/go-fastly/pull/481)
```
Integralist, line 9: The associated change is a breaking interface change so we need to be sure this is something we want to do now, and if so what other breaking changes can we roll up into the new major release.

PR #486, opened 2023-11-15
fix(resource): set ServiceVersion to an integer

fgsch:
Matches the type the resource API returns.


fgsch:
I've confirmed this across all the endpoints documented at https://developer.fastly.com/reference/api/services/resource/.

Integralist (approved):
LGTM. 

But would you be able to regenerate the fixtures just to be sure this works as expected. 

Thanks!

Integralist (approved):

PR #489, opened 2023-11-17
Replace most helpers with a generic ToPointer function

fgsch:
Discussed with @Integralist .


Integralist (approved):

fastly/basictypes_helper.go
```diff
@@ -1,27 +1,7 @@
...
-
-// Uint8 is a helper that returns a pointer to the uint8 value passed in.
-func Uint8(v uint8) *uint8 {
-	return &v
-}
-
-// Bool is a helper that returns a pointer to the bool value passed in.
-func Bool(v bool) *bool {
+// ToPointer converts T to *T.
+func ToPointer[T string | int | uint | uint8 | bool | CacheSettingAction | Compatibool | ERLAction | ERLLogger | ERLWindowSize | RequestSettingAction | RequestSettingXFF | S3AccessControlList | S3Redundancy | S3ServerSideEncryption](v T) *T {
```
joeshaw, line 4: Curious why you specify all these types rather than using `any`?

I wonder also if you could use the `~` sigil on the primitive types to get the underlying types?  If not `any` then `~string | ~int | ~uint | ~uint8 | ~bool`?

joeshaw: Haha never mind, I saw the follow-up PR.

PR #491, opened 2023-11-17
feat: generic paginator

Integralist:
This PR replaces the paginators for Services, ACL Entries and Dictionary Items with a generic paginator.

This PR does NOT replace the KV Stores Entries paginator as it has a completely different implementation.

Here is an example of how to consume the generic paginator...

> **NOTE:** I've included some code for custom sorting the returned API data (just in case the user wants to know how to do this themselves) but this code isn't necessary as the APIs already provide a field for sorting the data by a specific field (which again I demonstrate below).

```go
package main

import (
	"fmt"
	"log"
	"os"
	"sort"

	"github.com/fastly/go-fastly/v8/fastly" // uses `replace` directive in go.mod locally
)

const (
	serviceID = "<REDACTED>"
	aclID     = "<REDACTED>"
	dictID    = "<REDACTED>"
)

func main() {
	client, err := fastly.NewClient(os.Getenv("FASTLY_API_KEY"))
	if err != nil {
		log.Fatal(err)
	}

	p := client.GetServices(&fastly.GetServicesInput{
		Direction: "ascend",
		PerPage:   50,
		Sort:      "type",
	})

	var results []*fastly.Service
	for p.HasNext() {
		data, err := p.GetNext()
		if err != nil {
			fmt.Printf("failed to get next page (remaining: %d): %s", p.Remaining(), err)
			return
		}
		results = append(results, data...)
	}

	fmt.Printf("%#v\n", len(results))
	for _, service := range results {
		fmt.Printf("%#v (%s)\n", service.Name, service.Type)
	}

	sort.Stable(servicesByName(results))
	fmt.Printf("\nSorted:\n\n")
	for _, service := range results {
		fmt.Printf("%#v (%s)\n", service.Name, service.Type)
	}

	fmt.Printf("\nUsing ListServices:\n\n")
	listServiceResults, err := client.ListServices(&fastly.ListServicesInput{
		Direction: "ascend",
		Sort:      "type",
	})
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("(%d) %#v\n", len(listServiceResults), listServiceResults)

	fmt.Println("---------------------------------------------------------------")

	p2 := client.GetACLEntries(&fastly.GetACLEntriesInput{
		ACLID:     aclID,
		Direction: "ascend",
		PerPage:   50,
		ServiceID: serviceID,
		Sort:      "ip",
	})

	var results2 []*fastly.ACLEntry
	for p2.HasNext() {
		data, err := p2.GetNext()
		if err != nil {
			fmt.Printf("failed to get next page (remaining: %d): %s", p2.Remaining(), err)
			return
		}
		results2 = append(results2, data...)
	}

	fmt.Printf("%#v\n", len(results2))
	for _, aclEntry := range results2 {
		fmt.Printf("%s | %s\n", aclEntry.IP, aclEntry.ID)
	}

	sort.Stable(entriesByID(results2))
	fmt.Printf("\nSorted:\n\n")
	for _, aclEntry := range results2 {
		fmt.Printf("%s | %s\n", aclEntry.IP, aclEntry.ID)
	}

	fmt.Printf("\nUsing ListACLEntries:\n\n")
	listACLResults, err := client.ListACLEntries(&fastly.ListACLEntriesInput{
		ACLID:     aclID,
		Direction: "ascend",
		ServiceID: serviceID,
		Sort:      "ip",
	})
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("(%d) %#v\n", len(listACLResults), listACLResults)

	fmt.Println("---------------------------------------------------------------")

	p3 := client.GetDictionaryItems(&fastly.GetDictionaryItemsInput{
		DictionaryID: dictID,
		Direction:    "ascend",
		PerPage:      50,
		ServiceID:    serviceID,
		Sort:         "item_value",
	})

	var results3 []*fastly.DictionaryItem
	for p3.HasNext() {
		data, err := p3.GetNext()
		if err != nil {
			fmt.Printf("failed to get next page (remaining: %d): %s", p3.Remaining(), err)
			return
		}
		results3 = append(results3, data...)
	}

	fmt.Printf("%#v\n", len(results3))
	for _, dictItem := range results3 {
		fmt.Printf("%s | %s\n", dictItem.ItemKey, dictItem.ItemValue)
	}

	sort.Stable(dictionaryItemsByKey(results3))
	fmt.Printf("\nSorted:\n\n")
	for _, dictItem := range results3 {
		fmt.Printf("%s | %s\n", dictItem.ItemKey, dictItem.ItemValue)
	}

	fmt.Printf("\nUsing ListDictionaryItems:\n\n")
	listDictResults, err := client.ListDictionaryItems(&fastly.ListDictionaryItemsInput{
		DictionaryID: dictID,
		Direction:    "ascend",
		ServiceID:    serviceID,
		Sort:         "item_value",
	})
	if err != nil {
		log.Fatal(err)
	}
	fmt.Printf("(%d) %#v\n", len(listDictResults), listDictResults)

	fmt.Println("---------------------------------------------------------------")
}

// servicesByName is a list of services that are sortable by their name.
type servicesByName []*fastly.Service

// Len implement the sortable interface.
func (s servicesByName) Len() int {
	return len(s)
}

// Swap implement the sortable interface.
func (s servicesByName) Swap(i, j int) {
	s[i], s[j] = s[j], s[i]
}

// Less implement the sortable interface.
func (s servicesByName) Less(i, j int) bool {
	return s[i].Name < s[j].Name
}

// entriesByID is a list of ACL entries that are sortable by their ID.
type entriesByID []*fastly.ACLEntry

// Len implements the sortable interface.
func (s entriesByID) Len() int {
	return len(s)
}

// Swap implements the sortable interface.
func (s entriesByID) Swap(i, j int) {
	s[i], s[j] = s[j], s[i]
}

// Less implements the sortable interface.
func (s entriesByID) Less(i, j int) bool {
	return s[i].ID < s[j].ID
}

// dictionaryItemsByKey is a list of Dictionary items that are sortable by their key.
type dictionaryItemsByKey []*fastly.DictionaryItem

// Len implement the sortable interface.
func (s dictionaryItemsByKey) Len() int {
	return len(s)
}

// Swap implement the sortable interface.
func (s dictionaryItemsByKey) Swap(i, j int) {
	s[i], s[j] = s[j], s[i]
}

// Less implement the sortable interface.
func (s dictionaryItemsByKey) Less(i, j int) bool {
	return s[i].ItemKey < s[j].ItemKey
}
```


fastly/paginator.go
```diff
@@ -29,3 +37,178 @@ type PaginatorKVStoreEntries interface {
...
+	// Page is the current page.
+	Page int
+	// PerPage is the number of records per page.
+	PerPage int
+	// Sort is the field on which to sort.
+	Sort string
+}
+
+// Paginator represents a generic paginator.
+type Paginator[T ACLEntry | DictionaryItem | Service] interface {
```
joeshaw, line 61: Do you need this interface?  If nothing other than `ListPaginiator` implements it, it's probably not necessary.

fastly/paginator.go
```diff
@@ -29,3 +37,178 @@ type PaginatorKVStoreEntries interface {
...
+	HasNext() bool
+	Remaining() int
+	GetNext() ([]*T, error)
+}
+
+// ListPaginator implements the generic Paginator[N] interface.
+type ListPaginator[T ACLEntry | DictionaryItem | Service] struct {
```
joeshaw, line 68: I would probably have this take `any` rather than specific types.  What's the benefit to constraining it?

(Also, I think it's more idiomatic to constrain to an interface that contains the type sets, like `ordered` below.)

Integralist: Yup. Was originally `any` but code went (and still is) going through iterations. Pushed up as a draft as I'm having trouble figuring out how best to implement a generic sort as original code would use a stable sort that was unique fields per type. If you have thoughts on that I'd love to hear them 🙂

fastly/acl_entry.go
```diff
@@ -2,14 +2,15 @@ package fastly
...
 	"sort"
 	"strconv"
 	"time"
-
-	"github.com/peterhellberg/link"
 )
 
+// ACLEntriesPath is exposed primarily for use by the generic Paginator.
+// See ./paginator.go for details.
+const ACLEntriesPath = "/service/%s/acl/%s/entries"
```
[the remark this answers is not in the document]
Integralist: Thanks @joeshaw I'll review this tomorrow and decide on the follow-up 👍🏻

fastly/acl_entry.go
```diff
@@ -24,22 +19,12 @@ type ACLEntry struct {
...
+const aclEntriesPath = "/service/%s/acl/%s/entries"
 
-// Less implements the sortable interface.
-func (s entriesByID) Less(i, j int) bool {
-	return s[i].ID < s[j].ID
+// ACLEntriesPath is exposed primarily for use by the generic Paginator.
+// See ./paginator.go for details.
+func ACLEntriesPath(serviceID, aclID string) string {
+	return fmt.Sprintf(aclEntriesPath, serviceID, aclID)
 }
```
joeshaw, line 28: Given that you have `GetACLEntries` now, do you need to export this?  (Ditto elsewhere in the PR.)  A related question: do you need to export `NewPaginator`?  It looks like all uses of it are in the same package.

fastly/acl_entry.go
```diff
@@ -58,23 +43,14 @@ type ListACLEntriesInput struct {
...
-
-	return m
+// GetACLEntries returns a ListPaginator for paginating through the resources.
+func (c *Client) GetACLEntries(i *ListACLEntriesInput) *ListPaginator[ACLEntry] {
+	return NewPaginator[ACLEntry](c, &ListInput{
+		Direction: i.Direction,
+		Sort:      i.Sort,
+		Page:      i.Page,
+		PerPage:   i.PerPage,
+	}, ACLEntriesPath(i.ServiceID, i.ACLID))
```
joeshaw, line 53: It feels a little odd for `GetACLEntries` to take `ListACLEntriesInput` -- did you consider splitting these into separate types?

My initial thought was that you could have:

```go
type ListInput struct {
	Direction string
	Page int
	PerPage int
	Sort string
}

type ListACLEntriesInput {
	ACLID string
	ServiceID string
}

type GetACLEntriesInput {
	ListACLEntriesInput
	ListInput
}
```

but I think this doesn't quite work because `Direction` and `Sort` are relevant to `ListACLEntriesInput` too.

So I don't have a recommendation here, really.  I think you could probably construct a bunch of different types and compose them together here but I doubt it's worthwhile.  Curious to hear your thoughts on this.

fastly/paginator.go
```diff
@@ -1,31 +1,135 @@
...
-	Remaining() int
-	GetNext() ([]*DictionaryItem, error)
+// NOTE: We can't identify the underlying type of the type parameter T.
+// This is because we don't assign it to any of the defined function parameters.
+// If we did, then we could do this: https://go.dev/play/p/dfTMGjaSSAX.
+//
+// This means we have to have the caller pass the API path.
+// To make things easier we expose functions for constructing each path.
+// e.g. fastly.ServicePath(), fastly.ACLEntriesPath() etc.
+func NewPaginator[T any](client *Client, input *ListInput, path string) *ListPaginator[T] {
```
joeshaw, line 24: I mentioned this elsewhere but since the construction of a paginator is an implementation detail and all the uses of it are in the same package, you could unexport this.

joeshaw (commented):
Looks pretty good.  One suggestion and one question below.

fastly/paginator.go
Integralist: OK have made this private.

fastly/acl_entry.go
Integralist: Inlined this now.

fastly/acl_entry.go
Integralist: > It feels a little odd for GetACLEntries to take ListACLEntriesInput 

Yup. Agreed. I was going for the easy path of being able to pass the input for the List input directly into the Get.

I'll have a think about how to do this better.

Integralist: OK I opted for a small bit of duplication to give the GetN methods their own input.

fastly/acl_entry.go
```diff
@@ -47,6 +47,22 @@ func (c *Client) GetACLEntries(i *ListACLEntriesInput) *ListPaginator[ACLEntry]
...
+// ListACLEntriesInput is the input parameter to ListACLEntries function.
+type ListACLEntriesInput struct {
+	// ACLID is an alphanumeric string identifying a ACL (required).
+	ACLID string
+	// Direction is the direction in which to sort results.
+	Direction string
+	// Page is the current page.
+	Page int
+	// PerPage is the number of records per page.
+	PerPage int
```
joeshaw, line 59: These don't make sense for this endpoint, I think.

joeshaw (approved):
👍

PR #492, opened 2023-11-27
refactor(secret_store): replace json.NewEncoder with c.PostJSON

Integralist:


joeshaw (approved):
I suspect this was a holdover from when we had this implemented in an internal repo before it was moved here and had access to `PostJSON`.  @awilliams-fastly, do you know if this is the case?

awilliams-fastly (approved):
Yes, that sounds plausible. Or else I didn't know about `PostJSON` (looks like the TODO was added [later](#376))

PR #499, opened 2024-01-12
feat: fastly alerts

darin-nee:
This adds support for configuring and reviewing Fastly Alerts.

Fastly API docs (coming soon)

Further details: https://docs.fastly.com/en/guides/about-the-alerts-page


fastly/fastly_alerts.go
```diff
@@ -0,0 +1,344 @@
...
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strconv"
+	"time"
+)
+
+// AlertDefinition holds the configuration for one alert.
+type AlertDefinition struct {
+	ID                 string              `json:"id"`
```
Integralist, line 13: I'm a little undecided on the use of `json` struct tags vs `mapstructure` (https://github.com/mitchellh/mapstructure). 

On the one hand I feel like we should use mapstructure for consistency, because it is used currently throughout the entire codebase. 

But I also am not a fan of the 'magic' that is mapstructure 🙂 it's also an extra level of complexity in the code design that would be nice to avoid.

It's also never been overly clear to me why it was necessary to use it in the first place. I presume it's probably related to our API returning data types that aren't quite right for the field. For example, the API returns, what otherwise should be an integer, as a string. Using `mapstructure` means we're able to more easily marshal the data into the format we want it to be _exposed_ as (as part of the go-fastly interface) to the customer, rather than having to implement our own custom marshaler for each API endpoint.

But there has been a discussion internally at Fastly regarding the removal/replacement of mapstructure because it would help with the support of go-fastly on Fastly's Compute platform. So using native `json` unmarshaling here would help start us off in that direction.

So at the moment I'm in favour of using `json` but I might need to sit and dwell on it for a bit longer.

Integralist: OK having given this some thought I think it's OK to go ahead with `json` tags as I'm inclined to want to replace mapstructure in the future. We have an internal ticket tracking this 'replace mapstructure' request and I've commented there also.

fastly/fastly_alerts.go
```diff
@@ -0,0 +1,344 @@
...
+	// Name filters definitions by name substring (optional).
+	Name *string
+	// CreatedBy is for Fastly Staff use only (optional).
+	CreatedBy *string
+	// Cursor is the pagination cursor from a previous request's meta (optional).
+	Cursor *string
+	// Limit is the maximum number of items included in each response (optional).
+	Limit *int
+	// Sort is the field on which to sort definitions (optional).
+	Sort *string
```
Integralist, line 73: Please ensure all fields are sorted alphabetically. This makes it consistent with the rest of the codebase and also easier to search for fields (as a user). Please apply this request to **_all_** structs. Thanks!

fastly/fastly_alerts.go
```diff
@@ -0,0 +1,344 @@
...
+	// Metric is the name of the metric being monitored for alert evaluation (required).
+	Metric string `json:"metric"`
+	// Source is the metric source (required). Options are: 'stats', 'origins', 'domains'.
+	Source string `json:"source"`
+	// Dimensions are a list of origins or domains that the alert is restricted to.
+	Dimensions map[string][]string `json:"dimensions"`
+	// EvaluationStrategy is the evaluation strategy for the alert (required).
+	EvaluationStrategy map[string]any `json:"evaluation_strategy"`
+	// IntegrationIDs are IDs of integrations that notifications will be sent to.
+	IntegrationIDs []string `json:"integration_ids"`
```
Integralist, line 138: Please ensure all _input_ structs use pointers for their field types. If the type is a slice/map/channel then those don't need to be wrapped in a pointer as they're already pointer types (e.g. the Go runtime instantiates a [runtime.slice](https://github.com/golang/go/blob/883f062fc0a097bf561030ad453fd3e300896975/src/runtime/slice.go#L15), [runtime.hmap](https://github.com/golang/go/blob/883f062fc0a097bf561030ad453fd3e300896975/src/runtime/map.go#L117) or [runtime.hchan](https://github.com/golang/go/blob/883f062fc0a097bf561030ad453fd3e300896975/src/runtime/chan.go#L33C6-L33C11) and returns a pointer to them, so we can still carry out `nil` checks).

Integralist (changes requested):
Thanks @darin-nee this PR looks great. Just a couple of minor requested changes.

fastly/fastly_alerts.go
darin-nee: Thanks for the detailed explanation.

fastly/fastly_alerts.go
darin-nee: Fixed, for ***all*** structs.

fastly/fastly_alerts.go
darin-nee: Done.

darin-nee:
Thank you for the review @Integralist . I've made the revisions you suggested. I also removed a couple of optional parameters that had been included for "Fastly Staff use only" and one more that was mistakenly included in the first commit.

Integralist (approved):
Excellent work @darin-nee thank you! 

One of the better PRs I've seen in a while, much appreciated ❤️

PR #513, opened 2024-04-09
feat: notifications

darin-nee:
This adds support for configuring notifications.

Fastly API docs: https://www.fastly.com/documentation/reference/api/observability/notifications/

### DIFF
--- a/fastly/notifications.go
+++ b/fastly/notifications.go
@@ -10,14 +10,14 @@ import (
 
 // Integration holds the configuration for one integration.
 type Integration struct {
-	CreatedAt   *time.Time         `json:"created_at"`
-	Description *string            `json:"description"`
-	Config      *map[string]string `json:"config"`
-	ID          *string            `json:"id"`
-	Name        *string            `json:"name"`
-	Status      *string            `json:"status"`
-	Type        *string            `json:"type"`
-	UpdatedAt   *time.Time         `json:"updated_at"`
+	CreatedAt   *time.Time        `json:"created_at"`
+	Description *string           `json:"description"`
+	Config      map[string]string `json:"config"`
+	ID          *string           `json:"id"`
+	Name        *string           `json:"name"`
+	Status      *string           `json:"status"`
+	Type        *string           `json:"type"`
+	UpdatedAt   *time.Time        `json:"updated_at"`
 }
 
 // SearchIntegrationsInput is used as input to the SearchIntegrations function.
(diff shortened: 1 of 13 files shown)


fastly/notifications.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+// UpdateIntegrationInput is used as input to the UpdateIntegration function.
+type UpdateIntegrationInput struct {
+	// Config is configuration specific to the integration type.
+	Config map[string]string
+	// Description is the user submitted description of the integration.
+	Description *string
+	// ID of integration to update (required).
+	ID *string
```
Integralist, line 151: Typically anything `(required)` (which means it's used as part of the API request path) I've not made a pointer because it's not optional (i.e. we _need_ that input in order to actually make the API request).

Integralist (commented):
Thanks @darin-nee just a couple of comments that probably apply broadly.

fastly/notifications.go
```diff
@@ -0,0 +1,311 @@
...
+}
+
+// IntegrationsMeeta holds metadata about an integrations query.
+type IntegrationsMeta struct {
+	Limit      int    `json:"limit"`
+	NextCursor string `json:"next_cursor"`
+	Sort       string `json:"sort"`
+	Total      int    `json:"total"`
+	Type       string `json:"type"`
+}
```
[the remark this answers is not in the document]
darin-nee: Fixed, and thanks for the detailed explanation.

fastly/notifications.go
darin-nee: Got it, I've aligned the PR in this respect.

fastly/notifications.go
```diff
@@ -0,0 +1,311 @@
...
+	"net/http"
+	"strconv"
+	"time"
+)
+
+// Integration holds the configuration for one integration.
+type Integration struct {
+	CreatedAt   *time.Time         `json:"created_at"`
+	Description *string            `json:"description"`
+	Config      *map[string]string `json:"config"`
```
Integralist, line 15: We don't need the `*` for a `map` or slice type (e.g. `*[]Integration` defined below) as those types are already pointers (i.e. if you don't set a value for the field, then it will be `nil`).

Example:
https://play.golang.com/p/CddfmQajQ-N

Integralist (changes requested):
Thanks @darin-nee for addressing the feedback. I've just one remaining comment (if you could apply globally/wherever relevant). Thanks!

Integralist (approved):
LGTM. Thanks!

Integralist:
https://github.com/fastly/go-fastly/releases/tag/v9.3.0

PR #516, opened 2024-04-15
Add Image Optimizer default settings API

daboross:
Let me know if there's anything else I need here! I've got some corresponding PRs to other repositories.

I'm a bit unsure if my testing here is appropriate; it's half testing the library, but also half testing the endpoint behaves as I'd expect given the what the library is calling. I figure it should work, but I could also trim it down into a less verbose test.

### DIFF
--- a/fastly/image_optimizer_default_settings.go
+++ b/fastly/image_optimizer_default_settings.go
@@ -29,10 +29,15 @@ func (r ResizeFilter) MarshalJSON() ([]byte, error) {
 }
 
 const (
+	// A Lanczos filter with a kernel size of 3. Lanczos filters can detect edges and linear features within an image, providing the best possible reconstruction.
 	Lanczos3 ResizeFilter = iota
+	// A Lanczos filter with a kernel size of 2.
 	Lanczos2
+	// A filter using an average of a 4x4 environment of pixels, weighing the innermost pixels higher.
 	Bicubic
+	// A filter using an average of a 2x2 environment of pixels.
 	Bilinear
+	// A filter using the value of nearby translated pixel values. Preserves hard edges.
 	Nearest
 )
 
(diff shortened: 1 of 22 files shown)


fastly/image_optimizer_default_settings_test.go
```diff
@@ -0,0 +1,235 @@
+package fastly
+
+import (
+	"strings"
+	"testing"
+)
+
+// TestClient_ImageOptimizerDefaultSettings tests the Image Optimizer Default Settings API
+//
+// Note: to run this, all backends on the active version of the test service must have shielding
```
Integralist, line 10:
```suggestion
// NOTE: to run this, all backends on the active version of the test service must have shielding
```

Integralist (approved):
One small suggested edit, otherwise this looks fine to me. Thanks

Integralist:
You'll need to rebase `main` and fix the conflicting errors.

Also, we won't be able to merge this PR until the API documentation is live.

Integralist:
So can you move this PR back to "draft" until API documentation is ready. Just to avoid anyone accidentally merging. Thanks 🙂

Integralist:
@daboross you'll need to rebase `main` and address the conflict.

daboross:
Thanks for the review! I'll move this back to draft now, and rebase (and probably rebase again before merging!).

daboross:
I've merged the API documentation PR, now just waiting for that to go live.

fastly/image_optimizer_default_settings.go
```diff
@@ -0,0 +1,187 @@
...
+	return json.Marshal(r.String())
+}
+
+const (
+	// Match the input JPEG type, or baseline if transforming from a non-JPEG input.
+	Auto JpegType = iota
+	// Output baseline JPEG images
+	Baseline
+	// Output progressive JPEG images
+	Progressive
```
Integralist, line 69: These constants should be prefixed with `ImageOpto` as they will be globally available across the whole go-fastly package and in the future there might be other features/APIs that have things related to 'auto' or 'progressive' etc, so we need to be able to distinguish these types as being related to ImageOpto.

fastly/image_optimizer_default_settings.go
```diff
@@ -0,0 +1,187 @@
...
+	// A filter using an average of a 4x4 environment of pixels, weighing the innermost pixels higher.
+	Bicubic
+	// A filter using an average of a 2x2 environment of pixels.
+	Bilinear
+	// A filter using the value of nearby translated pixel values. Preserves hard edges.
+	Nearest
+)
+
+// JpegType is a base for different JpegType variants
+type JpegType int64
```
Integralist, line 45: This custom type should be prefixed with `ImageOpto` as it will be globally available across the whole go-fastly package and in the future there might be other features/APIs that have things related to jpegs (maybe?), so we need to be able to distinguish this type as being related to ImageOpto.

> **EDIT:** Let's consider this a general rule across this whole file as I realised there are a few different consts and types that could overlap with future requirements.

daboross:
API docs are live - https://www.fastly.chttps://www.fastly.com/documentation/reference/api/services/image-optimizer-default-settings/!

@Integralist Do you have any preference on how this is merged? I'd be happy to squash & merge this now but I want to check in first, since I know you'll be doing the release work & I don't want to presume.

fastly/image_optimizer_default_settings_test.go
```diff
@@ -0,0 +1,290 @@
...
+	defer func() {
+		record(t, fixtureBase+"final_reset", func(c *Client) {
+			if originalSettings != nil {
+				_, err = c.UpdateImageOptimizerDefaultSettings(&UpdateImageOptimizerDefaultSettingsInput{
+					ServiceID:      testServiceID,
+					ServiceVersion: *testVersion.Number,
+					// ResizeFilter:   &originalSettings.ResizeFilter,
+					Webp:        &originalSettings.Webp,
+					WebpQuality: &originalSettings.WebpQuality,
+					// JpegType:       &originalSettings.JpegType,
```
Integralist, line 71: Are these two comments meant to be deleted?

fastly/image_optimizer_default_settings.go
daboross: Right, of course! I do have a question - now that I'm looking for it, I see the other files doing this have the type name as a prefix to each constant - `ProductUndefined` for example in product_enablement.

Do you think `ResizeFilterAuto`, etc. (and `JpegTypeAuto`, etc.) could make sense, or would that still be too unspecific?

daboross: sounds good!

daboross: Should be fixed!

fastly/image_optimizer_default_settings_test.go
daboross: Fixed!

daboross:
Right - if the rest of this looks good, should I go ahead and merge this?

I think I asked that before but I want to ask again just to confirm :sweat_smile:

Integralist:
Just going to give the PR one last look over and then I'll re-approve so you can merge. Thanks!

daboross:
Thanks for the reviews & all the help here!

PR #517, opened 2024-04-16
Add support for specifying location when creating KV stores

hrmsk66:
:wave: This PR adds the ability to specify the location when creating a new KV store using [the API](https://www.fastly.com/documentation/reference/api/services/resources/kv-store/#create-store). I would like **terraform-provider-fastly** to support specifying the location for KV stores, and this change is a preparation for that.

Changes made:

- Added a new `Location` field to the `CreateKVStoreInput` struct.
- Updated the `CreateKVStore` method to include the location parameter in the API request if it is provided in the input struct.
- Added a new test `TestClient_CreateKVStoresWithLocations` to verify that KV stores can be created with different locations.

### DIFF
--- a/fastly/kv_store.go
+++ b/fastly/kv_store.go
@@ -24,7 +24,7 @@ type KVStore struct {
 type CreateKVStoreInput struct {
 	// Name is the name of the store to create (required).
 	Name string `json:"name"`
-	// Location is the location of the store (optional).
+	// Location is the regional location of the store (optional).
 	Location string `json:"-"`
 }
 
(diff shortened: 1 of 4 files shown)


fastly/kv_store.go
```diff
@@ -24,6 +24,8 @@ type KVStore struct {
 type CreateKVStoreInput struct {
 	// Name is the name of the store to create (required).
 	Name string `json:"name"`
+	// Location is the location of the store (optional).
```
Integralist, line 27:
```suggestion
	// Location is the regional location of the store (optional).
```

Integralist (approved):
LGTM but with one minor suggested edit.

Integralist:
https://github.com/fastly/go-fastly/releases/tag/v9.3.0

PR #519, opened 2024-04-17
fix(client): support tcp keepalive

Integralist:
Disabling TCP keepalive destroys application performance if it needs to make multiple requests to the Fastly API from across the globe (e.g. when deleting multiple keys from a KV Store).

This PR also adds some other small refactors to the client code (see [commits](https://github.com/fastly/go-fastly/pull/519/commits) for details).


phamann (approved):
LGTM

PR #521, opened 2024-04-17
Use constants to validate KV Store location values

hrmsk66:
This commit introduces the `KVStoreLocation` type to represent the regional location of KV stores. Constants are defined for the available KV store locations, and the `CreateKVStoreInput` struct and `CreateKVStore` method are updated to use the new type.

The test case `TestClient_CreateKVStoresWithLocations` is also modified to use the constants.


hrmsk66:
`Consistency` type handles it differently. Please let me know if it would be better to align with this approach.

https://github.com/fastly/go-fastly/blob/d375ad83be66dd32b32405a1730a21fc2b6b73d8/fastly/kv_store.go#L221-L238

PR #523, opened 2024-05-07
feat(tls_subscription): add `tls_certificate` attributes

Integralist:
This PR exposes additional metadata for a TLS Certificate associated with a TLS Subscription.

> **NOTE:** I've regenerated the test suite fixture files but I've not validated the change beyond that because I don't own the domains defined in the code, and consequently I haven't completed the DNS challenges required to move a TLS Subscription from "pending" to "issued".


gschorkopf (approved):
LGTM!

PR #527, opened 2024-05-20
Add ClonedFrom to PackageMetadata

jsocol:
We recently exposed the source repo for e.g. Compute Starter Kits via
the `cloned_from` key in the PackageMetadata resource in the API. This
makes it programmatically accessible.

### DIFF
--- a/fastly/compute_package.go
+++ b/fastly/compute_package.go
@@ -30,6 +30,7 @@ type PackageMetadata struct {
 	Language    *string  `mapstructure:"language"`
 	Name        *string  `mapstructure:"name"`
 	Size        *int64   `mapstructure:"size"`
+	ClonedFrom  *string  `mapstructure:"cloned_from"`
 }
 
 // GetPackageInput is used as input to the GetPackage function.


fastly/compute_package.go
```diff
@@ -30,6 +30,7 @@ type PackageMetadata struct {
 	Language    *string  `mapstructure:"language"`
 	Name        *string  `mapstructure:"name"`
 	Size        *int64   `mapstructure:"size"`
+	ClonedFrom  *string  `mapstructure:"cloned_from"`
```
Integralist, line 33: Can we sort the fields alphabetically (for code base consistency as well as making it easier for users to search for specific fields).

Integralist: It's not necessary I guess, but you could always regenerate the fixtures and update the tests to check for the existence of this field.

Integralist (approved):
Two minor comments, otherwise LGTM.

fastly/compute_package.go
jsocol: Done and done!

fastly/fixtures/package/get.yaml
```diff
@@ -7,11 +7,11 @@ interactions:
...
-      package template used by the Fastly CLI for Rust-based Compute@Edge projects.","authors":["fastly@fastly.com"],"language":"rust","size":2015936,"hashsum":"f99485bd301e23f028474d26d398da525de17a372ae9e7026891d7f85361d2540d14b3b091929c3f170eade573595e20b3405a9e29651ede59915f2e1652f616","files_hash":"a763d3c88968ebc17691900d3c14306762296df8e47a1c2d7661cee0e0c5aa6d4c082a7c128d6e719fe333b73b46fe3ae32694716ccd2efa21f5d9f049ceec6d"},"version":2}'
+    body: '{"created_at":"2023-05-11T10:57:54Z","id":"yiXgwy8Xl4n1jVMAbpM813","updated_at":"2023-05-11T10:57:54Z","service_id":"7i6HN3TK9wS159v2gPAZ8A","deleted_at":null,"metadata":{"name":"wasm-test","description":"Default
+    package template used by the Fastly CLI for Rust-based Compute@Edge projects.","authors":["fastly@fastly.com"],"language":"rust","size":2015936,"hashsum":"f99485bd301e23f028474d26d398da525de17a372ae9e7026891d7f85361d2540d14b3b091929c3f170eade573595e20b3405a9e29651ede59915f2e1652f616","files_hash":"a763d3c88968ebc17691900d3c14306762296df8e47a1c2d7661cee0e0c5aa6d4c082a7c128d6e719fe333b73b46fe3ae32694716ccd2efa21f5d9f049ceec6d","cloned_from":"https://github.com/foo/bar.git"},"version":2}'
```
Integralist, line 14: I'm guessing we have to manually edit the fixtures because the API itself isn't live or we don't have any real data to pull from the API?

jsocol: I certainly could pull real API data here, I didn't know that's where we'd generated this from. I admit I assumed it was meant to match the `valid.tar.gz` package in the test and didn't see a way to regenerate that.

Integralist: Yeah all the fixtures are generated from real API calls (see [docs here](https://github.com/fastly/go-fastly/blob/main/TESTING.md)). Unfortunately some times, depending on what fixture needs updating it's not just a case of deleting (for example...) the package fixture but also a service fixture, as you'll need a service first to upload a package into. Additionally the `valid.tar.gz` was likely generated using the CLI and then manually copied into the test folder (so you could update it that way).

jsocol: @Integralist done!

PR #529, opened 2024-06-04
add limit_workspaces to user model

nat-fastly:
The Fastly UI needs to read the `limit_workspaces` attribute from the `GET /user/:user_id` response in order to determine whether to display the list of workspaces on the user management screen for each user

### DIFF
--- a/fastly/user.go
+++ b/fastly/user.go
@@ -13,6 +13,7 @@ type User struct {
 	DeletedAt              *time.Time `mapstructure:"deleted_at"`
 	EmailHash              *string    `mapstructure:"email_hash"`
 	LimitServices          *bool      `mapstructure:"limit_services"`
+	LimitWorkspaces        *bool      `mapstructure:"limit_workspaces"`
 	Locked                 *bool      `mapstructure:"locked"`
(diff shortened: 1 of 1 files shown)


fastly/user.go
```diff
@@ -13,6 +13,7 @@ type User struct {
 	DeletedAt              *time.Time `mapstructure:"deleted_at"`
 	EmailHash              *string    `mapstructure:"email_hash"`
 	LimitServices          *bool      `mapstructure:"limit_services"`
+	LimitWorkspaces        *bool      `mapstructure:"limit_workspaces"`
```
Integralist, line 16: I was checking the customer API https://www.fastly.com/documentation/reference/api/account/customer/ and I don't see the `limit_workspaces` field documented. It's not a blocker for this PR but it's usually a good idea to make sure the docs are updated to reflect the latest interface.

Integralist (approved):
One minor comment, otherwise LGTM.

fastly/user.go
nat-fastly: @Integralist thank you for reviewing this! We're planning to update the API docs after we release this change

PR #531, opened 2024-06-05
feat(client): prepend custom user-agent via FASTLY_USER_AGENT env variable to user

jsonroy-fastly:
Added the use of a new environment variable that will allow the user to prepend values to the default User-Agent of the http client.


Integralist (approved):

Integralist (changes requested):
An internal discussion is required before this change can move forward

Integralist (approved):

sgorbaty (approved):

PR #542, opened 2024-08-27
feat(service_version): Add support for environments.

simonwistow:
Also Activating and Deactivating an Service Version to and from a particular Environment.

It's possible that we don't want the  StageVerson and UnstageVersion shortcuts.

### DIFF
--- a/fastly/service_version.go
+++ b/fastly/service_version.go
@@ -7,17 +7,25 @@ import (
 
 // Version represents a distinct configuration version.
 type Version struct {
-	Active    *bool      `mapstructure:"active"`
-	Comment   *string    `mapstructure:"comment"`
-	CreatedAt *time.Time `mapstructure:"created_at"`
-	DeletedAt *time.Time `mapstructure:"deleted_at"`
-	Deployed  *bool      `mapstructure:"deployed"`
-	Locked    *bool      `mapstructure:"locked"`
-	Number    *int       `mapstructure:"number"`
-	ServiceID *string    `mapstructure:"service_id"`
-	Staging   *bool      `mapstructure:"staging"`
-	Testing   *bool      `mapstructure:"testing"`
-	UpdatedAt *time.Time `mapstructure:"updated_at"`
+	Active       *bool          `mapstructure:"active"`
+	Comment      *string        `mapstructure:"comment"`
+	CreatedAt    *time.Time     `mapstructure:"created_at"`
+	DeletedAt    *time.Time     `mapstructure:"deleted_at"`
+	Deployed     *bool          `mapstructure:"deployed"`
+	Locked       *bool          `mapstructure:"locked"`
+	Number       *int           `mapstructure:"number"`
+	ServiceID    *string        `mapstructure:"service_id"`
+	Staging      *bool          `mapstructure:"staging"`
+	Testing      *bool          `mapstructure:"testing"`
+	UpdatedAt    *time.Time     `mapstructure:"updated_at"`
+	Environments []*Environment `mapstructure:"environments"`
+}
+
+// Environment represents a distinct deployment environment.
+type Environment struct {
+	ServiceVersion *int64  `mapstructure:"active_version"`
+	Name           *string `mapstructure:"name"`
+	ServiceID      *string `mapstructure:"service_id"`
 }
 
 // ListVersionsInput is the input to the ListVersions function.
@@ -176,6 +184,8 @@ type ActivateVersionInput struct {
 	ServiceID string
 	// ServiceVersion is the specific configuration version (required).
 	ServiceVersion int
+	// Environment is the Fastly environment to activate this version to (optional).
+	Environment string
 }
 
 // ActivateVersion activates the given version.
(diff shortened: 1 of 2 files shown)


Integralist:
👋🏻 

I don't see this API behaviour documented:
https://www.fastly.com/documentation/reference/api/services/version/#activate-service-version

What does the additional path segment do with regards to service activation?

kpfleming:
The API docs are still in progress.

fastly/service_version.go
```diff
@@ -174,6 +174,8 @@ type ActivateVersionInput struct {
 	ServiceID string
 	// ServiceVersion is the specific configuration version (required).
 	ServiceVersion int
+	// Environment is the fastly environment to activate this version to (optional)
```
kpfleming, line 177:
```suggestion
	// Environment is the Fastly environment to activate this version to (optional)
```

fastly/service_version.go
```diff
@@ -205,6 +212,8 @@ type DeactivateVersionInput struct {
 	ServiceID string
 	// ServiceVersion is the specific configuration version (required).
 	ServiceVersion int
+	// Environment is the fastly environment to deactivate this version from (optional)
```
kpfleming, line 215:
```suggestion
	// Environment is the Fastly environment to deactivate this version from (optional)
```

kpfleming (approved):
This looks fine to me other than properly capitalizing Fastly :-)

We'll wait to merge this until the API documentation has been merged, partially to ensure that the terminology ("Environment") matches up.

kpfleming:
I believe this is ready but asked for a second opinion :-)

fastly/service_version.go
```diff
@@ -329,3 +351,29 @@ func (c *Client) LockVersion(i *LockVersionInput) (*Version, error) {
...
+	// ServiceID is the ID of the service (required).
+	ServiceID string
+	// ServiceVersion is the specific configuration version (required).
+	ServiceVersion int
+}
+
+// UnstageVersion unstages the given version.
+func (c *Client) UnstageVersion(i *UnstageVersionInput) (*Version, error) {
+	return c.DeactivateVersion(&DeactivateVersionInput{ServiceID: i.ServiceID, ServiceVersion: i.ServiceVersion, Environment: "staging"})
+}
```
Integralist, line 379: This is an abstraction that I feel is unnecessary in the go-fastly API client.

We're adding 25 LOC to save the user from having to add _one_ additional field `Environment`.

**EDIT:** After saying that, I now recall the API currently only supports one _environment_, called "staging". So I presume we've added this abstraction to ensure users don't try to specify a different environment that would ultimately result in an API error.

The problem is a user can still stumble into issues because the `ActivateVersion` and `DeactivateVersion` methods are exposed publicly and are generally more known to users.

From a design perspective, we'll want to be careful exposing additional methods as we may have to support them long-term, and changes to expected inputs will require major version bumps and deprecation of methods. Also, users might expect more methods to be added in future when new environments are made available.

I would suggest instead that we delete these methods, and define an enum for the supported environments and set the enum as the type for the `Environment` variable. This will force the user to provide the enum that go-fastly exposes.

Something like the following:

> [!NOTE]
> Go's type system isn't great when it comes to enums 😒 
> See code comments for example of why

```go
package main

import (
	"fmt"
)

type Environment int64

func (e Environment) String() string {
	switch e {
	case EnvironmentStaging:
		return "staging"
	case EnvironmentProd:
		return "production"
	}
	return "unknown"
}

const (
	EnvironmentUndefined Environment = iota
	EnvironmentStaging
	EnvironmentProd
)

func activate(e Environment) {
	if e.String() == "unknown" {
		fmt.Println("Invalid environment:", e)
		return
	}
	fmt.Println("Activated environment:", e)
}

func main() {
	// THIS WORKS FINE
	activate(EnvironmentStaging)
	activate(EnvironmentProd)

	// THIS WILL FAIL TO COMPILE
	// "cannot use "staging" (untyped string constant) as Environment value in argument to activate"
	//activate("staging")

	// UNFORTUNATELY, THIS DOES ALSO WORK (THANKS GO :sigh:)
	// BUT IT AT LEAST MEANS THE USER HAS TO WORK HARDER TO SIDE STEP THE TYPE SYSTEM
	// AND THEY STILL CANT SPECIFY A NON-EXISTENT ENVIRONMENT
	activate(1) // i.e. staging
	activate(3) // Invalid environment: unknown
}
```

kpfleming: Indeed, this falls a bit into the 'validation before making an API call' situation, which we generally try to avoid.

As a simpler step, I would be fine with reverting ActivateVersion/DeactivateVersion to their original forms, and implementing some sort of private internal interface those functions (and StageVersion/UnstageVersion) use to actually make the API call. I don't think the enum machinery will end up being a good investment of our time, because today you aren't allowed to activate on 'production' (you must still use the undecorated 'activate' operation), and the potential for the list of environments to grow beyond 2 or 3 is very low.

kpfleming:
This has been simplified to remove the 'stage' and 'unstage' wrappers. It now only adds the optional 'Environment' parameter to 'Activate'/'Deactivate', and exposes the 'Environments' details included in API responses when service/version details are obtained.

Integralist (approved):

Issue #541, opened 2024-08-22
missing input validation leads to potential data exposure or unintended data manipulation


domcyrus:
The go-fastly library doesn't validate most inputs, allowing a bad actor to misuse the library and potentially gain access to data that you do not want to expose.

This simple example demonstrates how the `GetService` call can be altered into a `GetServiceDetails` call by manipulating the input.

```golang
package main

import (
	"fmt"

	"github.com/fastly/go-fastly/v9/fastly"
)

func main() {
	fc, err := fastly.NewClient("$MY_FASTLY_API_KEY")
	if err != nil {
		panic(err)
	}

	// maliscous service id
	serviceID := "$MY_FASTLY_SERVICE_ID/details?"

	service, err := fc.GetService(&fastly.GetServiceInput{
		ServiceID: serviceID,
	})
	if err != nil {
		panic(err)
	}
	fmt.Printf("Service: %+v\n", service)
}
```

Note that the output of the code does indeed show that GetServiceDetails is called but errors on the JSON unmarshal. Still, the error does contain most of the service details.

This vulnerability could allow attackers to craft inputs that invoke unintended API calls, leading to potential data exposure or manipulation.

domcyrus:
I believe the best solution would be to refactor the calls to RawRequest in a way that makes sure that path segments are safely escaped using url.PathEscape. This could be done by introducing a new function called SafeRawRequest with a different signature:
```golang
// SafeRawRequest accepts a verb, URL, and RequestOptions struct and returns the
// constructed http.Request and any errors that occurred.
func (c *Client) SafeRawRequest(verb string, pathSegments []string, ro *RequestOptions) (*http.Request, error) {
	// Ensure we have request options.
	if ro == nil {
		ro = new(RequestOptions)
	}

	safePaths := make([]string, len(pathSegments))
	for i, pathSegment := range pathSegments {
		safePaths[i] = url.PathEscape(pathSegment)
	}
	u := url.JoinPath(c.url.String(), safePaths...)

	// Existing logic for building and returning the request
	// ...
}
```

kpfleming:
We're struggling a bit to understand the nature of the potential exposure here. In the first sentence you've used the terms 'bad actor' and 'you', and based on my reading the 'bad actor' is someone who has received a Fastly API token and 'you' is the owner of that token.

If that's the case, then the 'bad actor' already has access to all content that is available to that token, regardless of whether they use go-fastly properly or improperly.

[...]

PR #543, opened 2024-08-29
fix: Ensure that service IDs provided to API operations are alphanumeric.

domcyrus:
Validates service id to be alphanumeric in all places where it is used.

Addresses #541


kpfleming:
I was debating about suggesting use of `unicode` instead of hard-coded comparisons, but I see that you already considered that and made the correct choice (only ASCII letters and digits should be accepted, not all Unicode letters and digits).

kpfleming (approved):

PR #544, opened 2024-09-06
fix: Ensure that all API endpoint URLs are constructed safely.

kpfleming:
#543 will be reverted before this PR is merged, as this achieves the same result (disallowing injection of path elements in serviceID fields).

As discussed in our chat today, this PR adds a `ToSafeURL` function which accepts a slice of strings that are used to construct the API endpoint URL. Each string is checked for unacceptable values (currently just `..`) and passed through `url.PathEscape`. The result is that if any components contains a 'special character' which could affect parsing or routing of the resulting URL, it will be URL-encoded for safety.

I attempted to find all places where API endpoint URLs were being constructed from one or more components dynamically, but it is certainly possible that I missed some!

While this PR touches a lot of files, I find that the code is more readable now, as the URL-construction process isn't polluted by `Sprintf` format strings.

### DIFF
--- a/fastly/helpers.go
+++ b/fastly/helpers.go
@@ -36,12 +36,16 @@ func NullString(v string) *string {
 
 // ToSafeURL produces a safe (no path traversal, no unsafe characters) URL
 // from the path components passed in.
+//
+// Unlike the normal behavior of url.JoinPath, this function skips
+// ".." components, ensuring that user-provided components cannot
+// remove code-provided components from the resulting path.
 func ToSafeURL(unsafeComponents ...string) string {
 	safeComponents := make([]string, len(unsafeComponents))
 
 	for i := range unsafeComponents {
-		if component := unsafeComponents[i]; component != ".." {
-			safeComponents[i] = url.PathEscape(component)
+		if unsafeComponents[i] != ".." {
+			safeComponents[i] = url.PathEscape(unsafeComponents[i])
 		}
 	}
 
--- a/fastly/helpers.go
+++ b/fastly/helpers.go
@@ -1,5 +1,7 @@
 package fastly
 
+import "net/url"
+
 // MultiConstraint is a generic constraint for ToPointer/ToValue.
 type MultiConstraint interface {
 	[]string | ~string | ~int | int32 | ~int64 | uint | uint8 | uint32 | uint64 | float64 | ~bool
@@ -30,4 +32,23 @@ func NullString(v string) *string {
 		return nil
 	}
 	return &v
+}
+
+// ToSafeURL produces a safe (no path traversal, no unsafe characters) URL
+// from the path components passed in.
+func ToSafeURL(components ...string) string {
+	for idx, component := range components {
+		if component == ".." {
+			components[idx] = ""
+		} else {
+			components[idx] = url.PathEscape(component)
+		}
+	}
+
+	// it is safe to ignore the error returned from JoinPath
+	// because the only time it will be non-nil is if parsing
+	// the base path fails, but that will not fail since it is
+	// a constant "/" string
+	result, _ := url.JoinPath("/", components...)
+	return result
 }
--- a/fastly/helpers.go
+++ b/fastly/helpers.go
@@ -36,19 +36,17 @@ func NullString(v string) *string {
 
 // ToSafeURL produces a safe (no path traversal, no unsafe characters) URL
 // from the path components passed in.
-func ToSafeURL(components ...string) string {
-	for idx, component := range components {
-		if component == ".." {
-			components[idx] = ""
-		} else {
-			components[idx] = url.PathEscape(component)
-		}
+func ToSafeURL(unsafeComponents ...string) string {
+	safeComponents := make([]string, len(unsafeComponents))
+
+	for i := range unsafeComponents {
+		safeComponents[i] = url.PathEscape(unsafeComponents[i])
 	}
 
 	// it is safe to ignore the error returned from JoinPath
 	// because the only time it will be non-nil is if parsing
 	// the base path fails, but that will not fail since it is
 	// a constant "/" string
-	result, _ := url.JoinPath("/", components...)
+	result, _ := url.JoinPath("/", safeComponents...)
 	return result
 }
--- a/fastly/helpers.go
+++ b/fastly/helpers.go
@@ -40,7 +40,9 @@ func ToSafeURL(unsafeComponents ...string) string {
 	safeComponents := make([]string, len(unsafeComponents))
 
 	for i := range unsafeComponents {
-		safeComponents[i] = url.PathEscape(unsafeComponents[i])
+		if component := unsafeComponents[i]; component != ".." {
+			safeComponents[i] = url.PathEscape(component)
+		}
 	}
 
 	// it is safe to ignore the error returned from JoinPath
--- a/fastly/stats_historical.go
+++ b/fastly/stats_historical.go
@@ -2,7 +2,6 @@ package fastly
 
 import (
 	"encoding/json"
-	"fmt"
 )
 
 // StatsResponse is a response from the service stats API endpoint.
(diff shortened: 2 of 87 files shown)


fastly/helpers.go
```diff
@@ -31,3 +33,18 @@ func NullString(v string) *string {
...
+func ToSafeURL(components ...string) string {
+	for idx, component := range components {
+		components[idx] = url.PathEscape(component)
+	}
+
+	// it is safe to ignore the error returned from JoinPath
+	// because the only time it will be non-nil is if parsing
+	// the base path fails, but that will not fail since it is
+	// a constant "/" string
+	result, _ := url.JoinPath("/", components...)
```
cee-dub, line 48: Underneath, this relies on https://pkg.go.dev/path#Join which preserves (but simplifies) `..` elements. Seems like that could still allow for path traversal?

kpfleming: Well, 'path traversal' in this case is very limited compared to an HTTP server that is serving content from a filesystem. `..` isn't going to have any useful effect in an API route being sent to our API gateway and/or Northstar.

The primary protection we're going for here is to handle `/` in data elements, and using `PathEscape` covers that. Using `JoinPath` was just easier than writing my own loop to combine all of the elements :)

cee-dub: FWIW, I think `JoinPath` is the right choice! We could do the easy thing and remove any elements that `== “..”` as well.

kpfleming: That would require building a new slice then, instead of modifying the one we have in place, since removing a component would modify the indices of the remaining components. That's probably better than the in-place modification anyway, we don't really need to optimize this path.

cee-dub: With path.Join, it will turn // into /, so an empty element is fine to replace any bad element.

kpfleming:
Also, I am fairly certain that all of these changes are correct, as during development when I made mistakes (more often than I wished) the existing tests caught those mistakes. Assuming we have 100% test coverage for the modified files, these changes should effectively be a no-op when proper data is provided to the API operations.

kpfleming:
If we wanted to get extra fancy we could improve `ToSafeURL` to receive `[]interface` instead of `[]string`, and then it could handle formatting the integers itself. That would eliminate most of the integer-formatting in the PR, except for one case where the resulting string has to be zero-padded.

fastly/stats_historical.go
```diff
@@ -147,14 +146,19 @@ func (c *Client) GetStatsField(i *GetStatsInput) (*StatsFieldResponse, error) {
 
 // GetStatsJSON fetches stats and decodes the response directly to the JSON struct dst.
 func (c *Client) GetStatsJSON(i *GetStatsInput, dst any) error {
-	p := "/stats"
+	components := make([]string, 1, 5)
```
cee-dub, line 149: Why 5? Looks like 3 total here.

fastly/stats_realtime.go
```diff
@@ -60,10 +60,11 @@ func (c *RTSClient) GetRealtimeStatsJSON(i *GetRealtimeStatsInput, dst any) erro
...
 
-	path := fmt.Sprintf("/v1/channel/%s/ts/%d", i.ServiceID, i.Timestamp)
-
+	var path string
 	if i.Limit != nil {
-		path = fmt.Sprintf("%s/limit/%d", path, *i.Limit)
+		path = ToSafeURL("v1", "channel", i.ServiceID, "ts", strconv.FormatUint(i.Timestamp, 10), "limit", strconv.FormatUint(uint64(*i.Limit), 10))
+	} else {
+		path = ToSafeURL("v1", "channel", i.ServiceID, "ts", strconv.FormatUint(i.Timestamp, 10))
 	}
```
cee-dub, line 68: Could easily do a components construction here:
```
components := []string{"v1", "channel", i.ServiceID, "ts", strconv.FormatUint(i.Timestamp, 10)}
if i.Limit != nil {
    components = append(components, "limit", strconv.FormatUint(uint64(*i.Limit), 10))
}
path := ToSafeURL(components)
```

cee-dub (approved):
Just some stylistic comments. Thanks for doing this! Lots of manual conversion.

fastly/helpers.go
```diff
@@ -31,3 +33,22 @@ func NullString(v string) *string {
...
+// ToSafeURL produces a safe (no path traversal, no unsafe characters) URL
+// from the path components passed in.
+func ToSafeURL(components ...string) string {
+	for idx, component := range components {
+		if component == ".." {
+			components[idx] = ""
+		} else {
+			components[idx] = url.PathEscape(component)
+		}
+	}
```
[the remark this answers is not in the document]
kpfleming: Thanks! I'm about to push a new version of this function anyway, since I'm not comfortable doing an in-place modification of the slice argument. Also, the docs for `JoinPath` explicitly say that `../` components will be removed, so the loop doesn't need to check for those (and `PathEscape` does not modify them).

fastly/stats_historical.go
kpfleming: It can be 5 if both `i.Service` and `i.Field` are non-nil.

kpfleming:
I've addressed the feedback and also added unit tests for the new helper function. Also, most of the conversion wasn't manual, it was done using some complicated `sed` substitutions :-)

fastly/helpers.go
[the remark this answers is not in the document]
kpfleming: I was incorrect; `JoinPath` does not remove `..` components, it resolves them, which isn't the behavior that we want. I've restored the test to ignore those components.

cee-dub:
> If we wanted to get extra fancy

I really like that it *isn’t* fancy.

fastly/helpers.go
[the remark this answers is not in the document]
cee-dub: Correct, they are “followed” if you imagine traversing the tree backwards. Important to check for them wherever untrusted input is expected.

I should have said so prior, but I looked at this pretty carefully for the review. Having tests that cover the expected edge cases is ideal, and comments on the function that it is critical to security and changes should be carefully vetted would be helpful.

fastly/helpers.go
```diff
@@ -31,3 +33,22 @@ func NullString(v string) *string {
...
 }
+
+// ToSafeURL produces a safe (no path traversal, no unsafe characters) URL
+// from the path components passed in.
+func ToSafeURL(unsafeComponents ...string) string {
+	safeComponents := make([]string, len(unsafeComponents))
+
+	for i := range unsafeComponents {
+		if component := unsafeComponents[i]; component != ".." {
+			safeComponents[i] = url.PathEscape(component)
```
cee-dub, line 44: The extra local variable `component` isn’t needed in Go, since a (modified) copy of the input is returned by `PathEscape`.  Perhaps a Rust idiom or habit showing up here?

fastly/helpers.go
```diff
@@ -31,3 +33,22 @@ func NullString(v string) *string {
...
 	return &v
 }
+
+// ToSafeURL produces a safe (no path traversal, no unsafe characters) URL
+// from the path components passed in.
+func ToSafeURL(unsafeComponents ...string) string {
+	safeComponents := make([]string, len(unsafeComponents))
+
+	for i := range unsafeComponents {
+		if component := unsafeComponents[i]; component != ".." {
```
cee-dub, line 43: Coming to this not knowing anything, I might wonder what would happen when we skip over this index. Probably best to comment that an empty component is ignored and the resulting path only contains single slashes between `safeComponents`.

cee-dub (approved):
Looking good!

fastly/helpers.go
cee-dub: I so see the test coverage, so that helps.

fastly/helpers.go
kpfleming: Old habits die hard :-) This was the "don't repeat array indexing" habit.

fastly/helpers.go
kpfleming: Comment added.

cee-dub (approved):
Merge this baby.

Integralist (approved):
LGTM. 🚀

PR #546, opened 2024-09-16
feat(custom_dashboards): Add support for Custom Dashboards APIs.

tlindsay:
[DI-1346](https://fastly.atlassian.net/browse/DI-1346) Implements client for [Observability Custom Dashboards API](https://www.fastly.com/documentation/reference/api/observability/custom-dashboards/)

[DI-1346]: https://fastly.atlassian.net/browse/DI-1346?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/fastly/observability_custom_dashboard.go
+++ b/fastly/observability_custom_dashboard.go
@@ -118,6 +118,64 @@ type VisualizationConfig struct {
 	PlotType PlotType `json:"plot_type"`
 }
 
+type dashboardItemOption interface {
+	apply(*DashboardItem)
+}
+type optionFunc func(*DashboardItem)
+
+func (f optionFunc) apply(di *DashboardItem) {
+	f(di)
+}
+
+func WithTitle(title string) dashboardItemOption {
+	return optionFunc(func(di *DashboardItem) {
+		di.Title = title
+	})
+}
+func WithSubtitle(subtitle string) dashboardItemOption {
+	return optionFunc(func(di *DashboardItem) {
+		di.Subtitle = subtitle
+	})
+}
+func WithSpan(span uint8) dashboardItemOption {
+	return optionFunc(func(di *DashboardItem) {
+		di.Span = span
+	})
+}
+func WithCalculationMethod(calculationMethod CalculationMethod) dashboardItemOption {
+	return optionFunc(func(di *DashboardItem) {
+		di.Visualization.Config.CalculationMethod = &calculationMethod
+	})
+}
+func WithFormat(format VisualizationFormat) dashboardItemOption {
+	return optionFunc(func(di *DashboardItem) {
+		di.Visualization.Config.Format = &format
+	})
+}
+
+func NewDashboardItem(sourceType DashboardSourceType, metrics []string, plotType PlotType, options ...dashboardItemOption) DashboardItem {
+	di := DashboardItem{
+		DataSource: DashboardDataSource{
+			Type: sourceType,
+			Config: DashboardSourceConfig{
+				Metrics: metrics,
+			},
+		},
+		Visualization: DashboardVisualization{
+			Type: VisualizationTypeChart,
+			Config: VisualizationConfig{
+				PlotType: plotType,
+			},
+		},
+	}
+
+	for _, o := range options {
+		o.apply(&di)
+	}
+
+	return di
+}
+
 type ListDashboardsResponse struct {
 	Data []ObservabilityCustomDashboard `json:"data"`
 	Meta DashboardMeta                  `json:"meta"`
--- a/fastly/observability_custom_dashboard.go
+++ b/fastly/observability_custom_dashboard.go
@@ -89,33 +89,33 @@ type PlotType string
 
 const (
 	PlotTypeLine         PlotType = "line"
-	PlotTypeBar                   = "bar"
-	PlotTypeDonut                 = "donut"
-	PlotTypeSingleMetric          = "single-metric"
+	PlotTypeBar          PlotType = "bar"
+	PlotTypeDonut        PlotType = "donut"
+	PlotTypeSingleMetric PlotType = "single-metric"
 )
 
 type Format string
 
 const (
 	FormatNumber       Format = "number"
-	FormatBytes               = "bytes"
-	FormatPercent             = "percent"
-	FormatRequests            = "requests"
-	FormatResponses           = "responses"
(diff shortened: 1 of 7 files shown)


fastly/observability_custom_dashboard.go
```diff
@@ -0,0 +1,275 @@
...
+
+type VisualizationConfig struct {
+	// CalculationMethod is the aggregation function to apply to the dataset (optional)
+	CalculationMethod *CalculationMethod `json:"calculation_method,omitempty"`
+	// Format indicates the unit used to format the data (optional, default: number)
+	Format *VisualizationFormat `json:"format,omitempty"`
+	// PlotType is the type of chart to display (required)
+	PlotType PlotType `json:"plot_type"`
+}
+
```
tlindsay, line 130: @kpfleming Do you think it's worth doing something like this to abstract-away the schema of the nested `DashboardItem` resources? If we supply a constructor, then the user doesn't need to know/care as much about how the data is modeled.

```suggestion
type DashboardItemOption interface {
	apply(*DashboardItem)
}
type optionFunc func(*DashboardItem)
func (f optionFunc) apply(di *DashboardItem) {
	f(di)
}

func WithTitle(title string) DashboardItemOption {
	return optionFunc(func(di *DashboardItem) {
		di.Title = title
	})
}
func WithSubtitle(subtitle string) DashboardItemOption {
	return optionFunc(func(di *DashboardItem) {
		di.Subtitle = subtitle
	})
}
func WithSpan(span uint8) DashboardItemOption {
	return optionFunc(func(di *DashboardItem) {
		di.Span = span
	})
}
func WithCalculationMethod(calculationMethod CalculationMethod) DashboardItemOption {
	return optionFunc(func(di *DashboardItem) {
		di.Visualization.Config.CalculationMethod = &calculationMethod
	})
}
func WithFormat(format VisualizationFormat) DashboardItemOption {
	return optionFunc(func(di *DashboardItem) {
		di.Visualization.Config.Format = &format
	})
}

func NewDashboardItem(sourceType DashboardSourceType, metrics []string, plotType PlotType, options ...DashboardItemOption) DashboardItem {
	di := DashboardItem{
		DataSource: DashboardDataSource{
			Type: sourceType,
			Config: DashboardSourceConfig{
				Metrics: metrics,
			},
		},
		Visualization: DashboardVisualization{
			Type: VisualizationTypeChart,
			Config: VisualizationConfig{
				PlotType: plotType,
			},
		},
	}

	for _, o := range options {
		o.apply(&di)
	}

	return di
}

```

fastly/observability_custom_dashboard.go
```diff
@@ -0,0 +1,275 @@
...
+const VisualizationTypeChart VisualizationType = "chart"
+
+type DashboardVisualization struct {
+	// Config describes configuration options for the given visualization (required)
+	Config VisualizationConfig `json:"config"`
+	// Type is type of visualization to display. Currently only "chart" is supported (required)
+	Type VisualizationType `json:"type"`
+}
+
+type PlotType string
```
tlindsay, line 89: @kpfleming I considered making these _types_ private, thereby forcing the user to use our exported constants, but I decided against it so that users aren't forced to upgrade the client in order to take advantage of future new functionality in the API.

fastly/observability_custom_dashboard.go
```diff
@@ -0,0 +1,275 @@
...
+	Span uint8 `json:"span"`
+	// Subtitle is a human-readable subtitle to display, often a description of the visualization (optional)
+	Subtitle string `json:"subtitle"`
+	// Title is a human-readable title to display (optional)
+	Title string `json:"title"`
+	// Visualization describes the way the DashboardItem should display data (required)
+	Visualization DashboardVisualization `json:"visualization"`
+}
+
+type DashboardSourceType string
```
tlindsay, line 56: Naming types/constants is really hard. I'm not sure how to balance user-ergonomics with collision-avoidance since all of our API wrappers live in the same package/namespace.

Advice/input is welcomed!

fastly/observability_custom_dashboard.go
```diff
@@ -0,0 +1,275 @@
...
+const (
+	ErrInvalidDataSourceType    = ""
+	ErrInvalidMetric            = ""
+	ErrInvalidVisualizationType = ""
+	ErrInvalidPlotType          = ""
+	ErrInvalidCalculationMethod = ""
+)
+
+// ObservabilityCustomDashboard is a named container for a custom dashboard configuration
+type ObservabilityCustomDashboard struct {
```
kpfleming, line 21: I'm a bit surprised that this struct doesn't have any `json` annotations.

fastly/observability_custom_dashboard.go
kpfleming: That might be useful... you could also consider implementing this using the Builder pattern, so that the invocation of `NewDashboardItem` doesn't get increasingly longer as more and more options are added.

fastly/observability_custom_dashboard.go
```diff
@@ -0,0 +1,275 @@
+package fastly
+
+import (
+	"encoding/json"
+	"net/http"
+	"strconv"
+	"time"
+)
+
+const path = "/observability/dashboards"
```
kpfleming, line 10: For consistency, I'd prefer to drop this and just use `ToSafeURL` everywhere. At some point in the future we'll probably improve ToSafeURL to accept a mixture of slices and strings, at which point this could be a 'const slice of strings' and passed as the first argument to ToSafeURL, but that day is not today.

kpfleming (changes requested):
Overall this looks pretty good, but I've asked for one relatively small change.

fastly/observability_custom_dashboard.go
ydnar: If you have generally acyclic type dependencies, it's a common pattern to split types/functions into a sub-package, eg `.../dashboard` which acts as a namespace.

The caller side looks like `dashboard.Options` etc.

ydnar: Go type aliases can be used to migrate gradually without breaking backwards compatibility.

ydnar (commented):
minor design notes

kpfleming (approved):

Issue #474, opened 2023-09-29
handle automation-tokens


domcyrus:
At the moment `go-fastly` doesn't handle `automation-tokens` https://developer.fastly.com/reference/api/auth-tokens/automation/. Add an way to manage them within the `go-fastly` package.

Integralist:
Thanks for opening this issue @domcyrus 

I'll create an internal ticket to track this work, as I agree we should have this feature supported.

kpfleming:
This was completed in #547.

PR #547, opened 2024-09-23
feat(automation_tokens): Add support for automation-tokens

Jesse-Hills:
This PR aims to add support for [Automation Tokens](https://www.fastly.com/documentation/reference/api/auth-tokens/automation/), as requested in [#474](https://github.com/fastly/go-fastly/issues/474).

There are a few things worth mentioning about this implementation, mostly due to the [Automation Tokens](https://www.fastly.com/documentation/reference/api/auth-tokens/automation/) docs being inaccurate.
- `DELETE` doesn't work for the `/automation-tokens` endpoint. The examples don't work as documented, and inspecting requests in the web UI using developer tools shows the web UI using the `/tokens` endpoint, even for automation tokens. (it appears to be an issue with the endpoint expecting a `customer_id` in the request, but there is no documentation on it being required or how to pass it in the request)
- The request/response data is inconsistent with the documentation and what is actually expected/returned from the API, some appear to return some form of JSON:API, while others return regular JSON.
- `/automation-tokens/{id}/services` is supposed to return a list of services associated with the provided token id, also doesn't appear to work, so I haven't included it in this PR, but the same can be achieved using `GetAutomationToken` and accessing the `Services` field of the returned `AutomationToken`.

### DIFF
--- a/fastly/automation_token.go
+++ b/fastly/automation_token.go
@@ -1,0 +1,185 @@
+package fastly
+
+import (
+	"fmt"
+	"net/http"
+	"time"
+)
+
+// AutomationTokenRole is used to match possible automation token roles.
+type AutomationTokenRole string
+
+const (
+	// BillingRole allows view access to basic information about service configurations,
+	// invoices, and account billing history.
+	BillingRole AutomationTokenRole = "billing"
+	// EngineerRole allows creating services and managing their configurations.
+	EngineerRole AutomationTokenRole = "engineer"
+	// UserRole allows view access to basic information about service configurations,
+	// and controls.
+	UserRole AutomationTokenRole = "user"
+)
+
+// AutomationTokenPaginator is used for pagination on AutomationToken endpoints.
+// as they return JSONAPI data.
+type AutomationTokenPaginator struct {
+	Data []*AutomationToken           `mapstructure:"data"`
+	Meta AutomationTokenPaginatorMeta `mapstructure:"meta"`
+}
+
+// AutomationTokenPaginatorMeta represents the metadata for an AutomationTokenPaginator.
+type AutomationTokenPaginatorMeta struct {
+	CurrentPage int `mapstructure:"current_page"`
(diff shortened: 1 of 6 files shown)


fastly/automation_token.go
```diff
@@ -0,0 +1,185 @@
...
+	Services []string `json:"services" url:"services,omitempty,brackets"`
+	// Username is the email of the user the token is assigned to.
+	Username *string `json:"-" url:"username,omitempty"`
+	// TLSAccess indicates whether TLS access is enabled for the token.
+	TLSAccess bool `json:"tls_access" url:"tls_access,omitempty"`
+}
+
+// CreateAutomationToken creates a new resource.
+func (c *Client) CreateAutomationToken(i *CreateAutomationTokenInput) (*AutomationToken, error) {
+	_, err := c.PostForm("/sudo", i, nil)
```
kpfleming, line 143: This was also copy-and-paste from the regular token code, but the doc for this function should indicate that `sudo` capability must be available for the token that is used for this function.

kpfleming (approved):
Just some minor doc comments.

fastly/automation_token.go
```diff
@@ -0,0 +1,185 @@
...
+	if i.Page != nil {
+		input.Page = *i.Page
+	}
+	if i.PerPage != nil {
+		input.PerPage = *i.PerPage
+	}
+	return NewPaginator[AutomationTokenPaginator](c, input, "/automation-tokens")
+}
+
+// ListAutomationTokens retrieves all resources. Not suitable for large collections.
```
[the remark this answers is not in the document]
Jesse-Hills: This was copy-and-paste. I cannot qualify what 'large' is for this endpoint, so I have removed that note from the comment as it wasn't from the regular token code, and they are almost identical structs.

fastly/automation_token.go
Jesse-Hills: This has been updated to include a note about `sudo` being required.

PR #550, opened 2024-09-30
feat(product_enablement): Add support for NGWAF.

kpfleming:


philippschulte (commented):
This looks good to me but according to https://www.fastly.com/documentation/reference/api/products/enablement/ there are two more endpoints. Would it not make sense to add those to the test suite?

kpfleming:
Yes it would, but that will need to be done as a separate task I believe. I'll enter an issue for that.

kpfleming:
I've created (internal) issue CDTOOL-932 for adding the product configuration endpoints.

philippschulte (approved):
Thank you!

Issue #548, opened 2024-09-30
add origin_fetch_resp_*_bytes in Stats struct


kartheekkaipa:
Currently go-fastly do not have field for bytes fetched from origin for Stats response for GetStatsField function.

- origin_fetch_resp_body_bytes
- origin_fetch_resp_header_bytes

[fastly/stats_historical.go](https://github.com/fastly/go-fastly/blob/main/fastly/stats_historical.go)

if it's okay, I will raise the PR for the change.

kpfleming:
Indeed, those are documented [here](https://github.com/fastly/api-documentation/blob/7a30e055454d200cfa259b5e09c9a0567cbe08c7/src/historical.yaml#L527) and [here](https://github.com/fastly/api-documentation/blob/7a30e055454d200cfa259b5e09c9a0567cbe08c7/src/historical.yaml#L530). A PR would be very welcome!

kartheekkaipa:
I will start with changes.

I am getting 404 for the above links, can you please check the same.

Thank you

kpfleming:
Oh, sorry, those were internal links, my apologies!

The public documentation for those fields is [here](https://www.fastly.com/documentation/reference/api/metrics-stats/historical-stats/).

Take a look at the [CONTRIBUTING](https://github.com/fastly/go-fastly/blob/main/CONTRIBUTING.md) and [TESTING](https://github.com/fastly/go-fastly/blob/main/TESTING.md) files to understand what you'll need to do to enhance the existing [test fixtures](https://github.com/fastly/go-fastly/tree/main/fastly/fixtures/stats) to support those new fields.

Jelmund:
I opened a PR for this issue at: https://github.com/fastly/go-fastly/pull/552

PR #552, opened 2024-10-02
feat(stats_historical): add origin statistics

Jelmund:
Fixes https://github.com/fastly/go-fastly/issues/548


kpfleming:
This looks great! While we're in here, would you be willing to add all of the other `origin` stats that are not currently included?

kpfleming:
Thanks for the contribution!

Jelmund:
My pleasure!

PR #553, opened 2024-10-04
feat(stats_historical): add fields with all prefix

Jelmund:
This PR adds all fields with the "all" prefix to the stats_historical.go Stats struct.


Integralist (approved):
Thanks for opening this PR 👍🏻 🙂

Integralist:
GitHub thinks there's a conflict? Maybe if you rebase `main` it might correct itself (as I don't understand why it thinks there is a conflict to be resolved because all the UI shows me are the new fields you've added 🤷🏻)

Jelmund:
Yeah, I was also confused by this, but rebasing seems to have solved it! Thanks for the help. By the way, would it be possible to get an indication of when the next release might be expected?

Integralist:
Thanks @Jelmund 

re:

> By the way, would it be possible to get an indication of when the next release might be expected?

I'll have to let @kpfleming comment on that.

Jelmund:
I added another field in this PR, if you don't mind.

Integralist (approved):

kpfleming:
There is no fixed release schedule; we typically make releases when we've added functionality that is needed by the Fastly CLI or Fastly Terraform Provider, which is what happened this week.

There are some other additions to go-fastly coming in the next week or so, so I'll probably make a release once those are in place.

PR #555, opened 2024-10-14
Allow to manage default cert for custom TLS configurations

jenskueper:
This PR allows to manage the default certificate for custom TLS configuration. This is relevant for PlatformTLS.

### DIFF
--- a/fastly/fixtures/custom_tls_configuration/get.yaml
+++ b/fastly/fixtures/custom_tls_configuration/get.yaml
@@ -12,7 +12,7 @@ interactions:
     url: https://api.fastly.com/tls/configurations/TLS_CONFIGURATION_ID
     method: GET
   response:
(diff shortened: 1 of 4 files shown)


kpfleming:
After some internal discussion it's going to be slightly challenging to accept this PR. While the code for the API endpoint is fine, the tests are not, because they will fail unless they are run using a user account that has one of the (limited) products enabled which support this field.

I think we'll need to have a separate set of test fixtures and test cases, which are skipped unless an environment variable named `TEST_TLS_DEFAULT_CERTFICATE` is set.

CC @Integralist

Integralist:
The tests are run initially against the live API and those responses are saved into fixture files. 

If the fixtures don't need updating any time soon then maybe it's fine. But yes, upgrading will require someone with access to this feature.

jenskueper:
Hey, sorry for taking so long to get back to you. Should I modify the test case to only run conditionally or do you want to update the fixtures on your side?

kpfleming:
No problem; I'd prefer to see a new set of test cases/fixtures which are only run when a suitable environment variable is set to trigger them.

PR #559, opened 2024-10-31
[LOG-2040] Fix Grafana errors

mdenomy:
Fixed errors found during cli dev

* Pluralization of list command
* Added MessageType, User, and URL fields

### DIFF
--- a/fastly/logging_grafanacloudlogs.go
+++ b/fastly/logging_grafanacloudlogs.go
@@ -78,7 +78,7 @@ type CreateGrafanaCloudLogsInput struct {
 	Index *string `url:"index,omitempty"`
 	// Token is the API key from your GrafanaCloudLogs account.
 	Token *string `url:"token,omitempty"`
-	// Index
+	// Grafana User ID
 	User *string `url:"user,omitempty"`
 	// URL is the URL to stream logs to. Must use HTTPS.
 	URL *string `url:"url,omitempty"`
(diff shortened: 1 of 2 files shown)


fastly/logging_grafanacloudlogs.go
```diff
@@ -24,16 +25,16 @@ type GrafanaCloudLogs struct {
 	User  *string `mapstructure:"user"`
 }
 
-// ListGrafanaCloudLogsInput is used as input to the ListGrafanaCloudLogs function.
-type ListGrafanaCloudLogsInput struct {
+// ListGrafanaCloudLogssInput is used as input to the ListGrafanaCloudLogs function.
```
kpfleming, line 28: This looks incorrect, it's got a double 's'.

mdenomy: Yeah, it does look incorrect, but that seems to be the convention that the ListInput variables use for the logging endpoints, e.g. `ListGCSsInput` for GCS and `ListKafkasInput` for Kafka. It looks really weird if the endpoint type ends in an s, like GCS. I don't know if it is 100% consistent, some don't seem to have it, like Logentries, but that seemed to be the convention. Happy to change it if you think it reads better. I originally had it with a single s, so 🤷 

See https://github.com/fastly/go-fastly/blob/main/fastly/logging_gcs.go#L34 and https://github.com/fastly/go-fastly/blob/main/fastly/logging_kafka.go#L37

kpfleming: I think in this situation it comes down to whether the noun is singular or plural. GCS is 'Google Cloud Storage', so `GCSs` is "Google Cloud Storages" :-) Since "Logs" is already plural, we don't need the additional 's'.

mdenomy: Resolved in https://github.com/fastly/go-fastly/pull/559/commits/460b99554827be8fcb3a3259fe41979720e3ea43

kpfleming (approved):

PR #561, opened 2024-11-05
fix(debug_mode): Fix FASTLY_DEBUG_MODE when used in combination with go-vcr.

kpfleming:
While running tests, if go-vcr was enabled (VCR_DISABLE was not set to '1'), and FASTLY_DEBUG_MODE was set to 'true', and no go-vcr cassettes existed for the tests being run, then any tests which sent HTTP request with bodies (PUT, PATCH, etc.) would fail as the combination of these features would result in an empty body being sent as part of the request.

This was discovered while running tests for new endpoints, so no cassettes existed, and FASTLY_DEBUG_MODE was enabled in order to monitor the actual HTTP requests and responses.

### DIFF
--- a/fastly/client.go
+++ b/fastly/client.go
@@ -324,8 +324,22 @@ func (c *Client) Request(verb, p string, ro *RequestOptions) (*http.Response, er
 
 	if c.DebugMode {
 		r := req.Clone(context.Background())
+		// 'r' and 'req' both have a reference to a Body that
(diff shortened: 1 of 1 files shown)


kpfleming:
CC @Integralist 

This was a nasty one, I spent nearly two hours today trying to figure out why I couldn't get tests to pass...

Integralist (approved):
Wow this must have been frustrating. Good spot 👍

PR #562, opened 2024-11-06
test(infrastructure): Add support for testing against both Delivery and Compute services.

kpfleming:
Some API endpoints, or specific values provided to API endpoints, are only permitted for one service type or the other. This PR adds support for test SIDs for both Delivery and Compute services, updates the default SIDs to the ones recently created and owned by the DevEx Engineering team, and updates TESTING.md to document how they are used. There are no tests in the PR which make use of the Compute SID, but that will be included in a forthcoming commit.

Note that the test fixtures *were* not regenerated to include the new SIDs; regenerating all the test fixtures will take some time.


philippschulte (approved):
LGTM. Thanks

PR #563, opened 2024-11-06
test(product_enablement): Add test suites for all supported products.

kpfleming:
All supported products now have test suites which test 'get',
'enable', and 'disable' functionality. This requires using a Compute
service for testing as the Fanout product can only be enabled on
Compute services.

Additionally some small improvements were made in the core product
enablement code: unnecessary usage of 'fmt' was removed, and the
'enable' operation now sends the API call with a JSON body instead of
a www-form-urlencoded body.


philippschulte (approved):
Looking good. Thanks!

PR #567, opened 2024-11-26
POC for generating API endpoints using code generation, take 2

kpfleming:
This PR is a second proof-of-concept for generating API endpoints using the Jennifer code-generation module. It generates endpoints for Brotli compression and NGWAF as examples. For NGWAF, it includes configuration-during-enablement of the 'workspace_id', and configuration of 'workspace_id' and 'traffic_ramp' after enablement.


Makefile
```diff
@@ -46,11 +41,49 @@ dev-dependencies: ## Downloads the necessary dev dependencies.
...
+tidy: ## Cleans the Go module.
+	@echo "==> Tidying module"
+	@$(GO) mod tidy
+.PHONY: tidy
+
+generate: ## Builds and runs generators, to generate code for API endpoints and tests.
+	@echo "==> Removing existing generated code"
+	@rm -f fastly/products/*/api.go fastly/products/*/api_test.go
+	@echo "==> Building generators"
+	@$(GO) build -o generators ./internal/generators/...
```
ydnar, line 53: Why build first? Can this just be go run?

Second, could these be embedded as go:generate lines?

fastly/client.go
```diff
@@ -651,8 +651,8 @@ func checkResp(resp *http.Response, err error) (*http.Response, error) {
 	}
 }
 
-// decodeBodyMap is used to decode an HTTP response body into a mapstructure struct.
-func decodeBodyMap(body io.Reader, out any) error {
+// DecodeBodyMap is used to decode an HTTP response body into a mapstructure struct.
+func DecodeBodyMap(body io.Reader, out any) error {
```
ydnar, line 655: Can you have the exported func call the non-exported func, to minimize the size of this diff?

Makefile
kpfleming: Using `go run` in the `go:generate` lines would require that those lines have the relative path to the generator source code... that seemed a bit icky and fragile to me. This method produces binaries that are temporarily in the `PATH`, so the `go:generate` lines only need to include the name of the desired generator.

fastly/client.go
kpfleming: I could... or I could put this change to expose `decodeBodyMap` into a separate PR.

kpfleming: This change (and other related changes) have now been moved to their own commit, which I'm maintaining in a local branch. Once this is ready, I'll push that set of changes in their own PR first.

kpfleming:
Closing this as I've opted for a radically different approach which does not use code generation.

PR #569, opened 2024-12-07
refactor: Expose various test types/functions/variables for other packages.

kpfleming:
To support future API endpoints being located outside of the 'fastly' package, a number of types/functions/variables used in the implementations of those endpoints (and their unit tests) need to be exposed from the 'fastly' package.

In addition, the 'fastly_test.go' file must be given a different name, as files named '..._test.go' cannot expose any of their declarations outside of their package.

Finally, add support for Next-Gen WAF workspace IDs along side Delivery and Compute service IDs, as future API endpoints will require them.

### DIFF
--- a/fastly/fastly_test_utils.go
+++ b/fastly/fastly_test_utils.go
@@ -10,23 +10,29 @@ import (
 	"github.com/dnaeon/go-vcr/recorder"
 )
 
-// testClient is the test client.
-var testClient = DefaultClient()
+// TestClient is the test client.
+var TestClient = DefaultClient()
 
-// testStatsClient is the test client for realtime stats.
-var testStatsClient = NewRealtimeStatsClient()
+// TestStatsClient is the test client for realtime stats.
+var TestStatsClient = NewRealtimeStatsClient()
 
 // ID of the Delivery service for testing.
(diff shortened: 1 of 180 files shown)


kpfleming:
The relevant changes are in `Makefile`, `TESTING.md`, `client.go`, and `fastly_test_utils.go`. The changes in all the other files are only search-and-replace changes because the case of the names they reference changed.

fastly/fastly_test_utils.go
```diff
@@ -55,13 +61,21 @@ func computeServiceIDForTest() string {
 	return defaultComputeTestServiceID
 }
 
+func ngwafWorkspaceIDForTest() string {
+	if tsid := os.Getenv("FASTLY_TEST_NGWAF_WORKSPACE_ID"); tsid != "" {
+		return tsid
+	}
+
```
Integralist, line 68: lol, this is such a nit I'm almost embarrassed to comment 😅 but could we lose the line break between the the if statement and the return as we don't need the breathing space (IMHO).
```suggestion
```

Integralist (approved):
I looked at the specific files called out in the PR description and then scanned the rest (and they all looked to be in order).

PR #572, opened 2024-12-17
fix(products): Improve API usability.

kpfleming:
* Add 'EnableOutput' type aliases in each product, so that users of the API don't need to know whether that product has its own EnableOutput or not. Add 'NewEnableOutput' function so that users of the API can easily construct mock output instances.

* Add 'ProductName' constant in each product, so that users of the API will have access to the canonical name of the product along with its product_id.

* Change package names to conform to Go guidelines (remove underscores).

* Change APIs to accept and return by-value instead of by-pointer, as using by-pointer structures can be difficult in generic parameter contexts and the structures involved are small (so there is no real benefit to avoiding by-value passing).


cee-dub (commented):
I really like the package restructuring and the `const`s for product names and IDs. This clarifies things and makes it easier to use this correctly.

I'm not (presently, maybe I'll change my mind?) a fan of the type aliases. It makes it seem like there are a bunch of different/new types when they're actually almost all the same `productcore.EnableOutput`.

kpfleming:
The motivation for that is when a consumer of `bot_management.Enable` needs to know the type of the output (they won't always, but for testing they likely do), they would need to know whether that type is `products.EnableOutput` or `bot_management.EnableOutput`, and if that changed in the future it would be a breaking change. Aliasing to `bot_management.EnableOutput` addresses that problem well.

kpfleming:
The `ProductOutput` and related types have moved *back* to the `fastly/products` package, as they are going to be useful in the CLI as a consumer of these APIs.

fastly/products/botmanagement/api.go
```diff
@@ -1,25 +1,30 @@
...
 	"github.com/fastly/go-fastly/v9/internal/productcore"
 )
 
-const ProductID = "bot_management"
+const (
+	ProductID   = "bot_management"
+	ProductName = "Bot Management"
+)
+
+type EnableOutput = products.EnableOutput
```
Integralist, line 14: Maybe for future clarity, we could add an annotation like the following to all instances of this type of aliasing...

```suggestion
// EnableOutput distinguishes the product type from the general form.
type EnableOutput = products.EnableOutput
```

fastly/products/logexplorerinsights/api.go
```diff
@@ -1,25 +1,30 @@
-package log_explorer_insights
+package logexplorerinsights
```
Integralist, line 1: **Random question:** are these packages already in a public release? And if so, does renaming them result in a breaking change? As people might be relying on the naming of these public packages.

Integralist (approved):

fastly/products/logexplorerinsights/api.go
kpfleming: No, they have not been included in a release yet.

fastly/products/botmanagement/api.go
kpfleming: This is a good idea, I'll do that now so I won't forget.

PR #588, opened 2025-01-24
build(actions): Update GitHub Actions.

kpfleming:
Also enabled Dependabot automatic updates for Actions dependencies.


kpfleming:
The new version of `actions/setup-go` handles cache storage and retrieval of the Go modules directory itself, so that is why the `actions/cache` invocations were removed in this PR. Looking at the CI jobs, that appears to be working.

anthony-gomez-fastly (approved):

Issue #594, opened 2025-02-05
Compute ACL lookup call doesn't handle 2xx — Success response codes properly


philippschulte:
The compute ACL lookup [endpoint](https://www.fastly.com/documentation/reference/api/acls/acls/#compute-acl-lookup-acls) can return multiple different 2xx status codes and the go-fastly client doesn't handle them properly.

On a success the API is responding with a status code `200 - OK`
If the lookup IP is not part of the ACL then the API is responding with a status `204 - NO CONTENT`.

Since the go-fastly client is not returning any of the response information to handle multiple different 2xx status codes I would recommend to treat a status `204 - NO CONTENT` as an error.

This is a bug because if the API responds with a `204 - NO CONTENT` then the go-fastly client is unable to decode the json response because it doesn't exist: `failed to decode json response: EOF`

PR #595, opened 2025-02-06
Fix compute ACL lookup call for status 204

philippschulte:
Fix for #594.
 
 All Submissions:

### Changes to Core Features:


fastly/computeacls/api_lookup.go
```diff
@@ -33,6 +34,10 @@ func Lookup(c *fastly.Client, i *LookupInput) (*ComputeACLEntry, error) {
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("204 - No Content")
```
philippschulte, line 38: I didn't use `resp.Status` because that doesn't include a `-` and we usually format HTTP errors like `status - message`.

kpfleming (dismissed):
This looks fine to me, although I think we should communicate to the API owners that returning '204 No Content' when the requested IP is not present in the ACL is not terribly logical. '404 Not Found' would make a lot more sense.

fastly/computeacls/api_lookup.go
SahibYar:
```
fmt.Errorf("%d - %s", resp.StatusCode, http.StatusText(resp.StatusCode))
```

fastly/computeacls/api_test.go
```diff
@@ -194,6 +194,25 @@ func TestClient_ComputeACL(t *testing.T) {
 			page++
 		}
 	})
+
+	// Lookup a non-existing IP in the test compute ACL
+	fastly.Record(t, "lookup_non_existing_ip", func(c *fastly.Client) {
+		ip, ipNet, err := net.ParseCIDR("73.49.184.42/24")
+		if err != nil {
+			t.Errorf("error parsing IP: %v", err)
```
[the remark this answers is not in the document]
philippschulte: Addressed in d915bb1294f5fdb4f81f60cff58c792bcd025695.

fastly/computeacls/api_lookup.go
philippschulte: Addressed in a2c12c8ff69dbb665b1ea6e138a98da4924582c0.

fastly/computeacls/api_test.go
```diff
@@ -194,6 +194,25 @@ func TestClient_ComputeACL(t *testing.T) {
 			page++
 		}
 	})
+
+	// Lookup a non-existing IP in the test compute ACL
+	fastly.Record(t, "lookup_non_existing_ip", func(c *fastly.Client) {
+		ip, ipNet, err := net.ParseCIDR("73.49.184.42/24")
```
kpfleming, line 200: According to the API docs, the lookup operation accepts an address, not an address/mask combination. Given that, I don't think there's any need to use `ParseCIDR` here, as the test value should just be `73.49.184.42`.

kpfleming (approved):
LGTM

fastly/computeacls/api_lookup.go
```diff
@@ -33,6 +34,10 @@ func Lookup(c *fastly.Client, i *LookupInput) (*ComputeACLEntry, error) {
 	}
 	defer resp.Body.Close()
 
+	if resp.StatusCode != http.StatusOK {
```
awilliams-fastly, line 37: In the case that no matching IP was found, the API will return a `204 No Content`. This is not an error condition, rather a lack of results.

Instead of returning an error here, consider alternatives to allow the caller to distinguish between an error (e.g. network timeout) and this case. One idea is to return `(nil, nil)`.

PR #596, opened 2025-02-06
Allow untrusted root option on CreateCustomTLSCertificateInput

jdw6359:
# Details

The fastly api for creating / uploading custom TLS certificates allows for passing in the `allow_untrusted_root` flag, indicating that a self signed CA (untrusted) is permitted to sign the server auth certificates that get uploaded.

The go-fastly client does not support this flag, presumably because exposing this option could potentially lead to negative consequences (someone uploading a self signed cert that browsers don't trust for a domain that is already seeing production traffic).

Without this change, we (SeatGeek) am unable to have a fully localized test environment leveraging a self signed CA that I create on the fly to test / validate our integration with the fastly custom tls cert upload process.

## Rationale

* This attribute `allow_untrusted_root` is available on the API, so it should also be present on the official sdk.
* This attribute is off by default, with a full description of the drawbacks to using this flag.
* Even if clients were to set this flag, **this feature is still gated behind an account level flag** that must be turned on by someone internal to Fastly.

Given the multiple layers of protection and the "default off" behavior of this flag, I think that this attribute can be safely reflected in the schema without consequence to customers

Without this, we wont be able to leverage the go-fastly client which would be a bummer.

All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

This change is backwards compatible with existing behavior.

This change will allow clients to enable `allow_untrusted_root` when uploading a custom TLS cert

None

### DIFF
--- a/fastly/tls_custom_certificate.go
+++ b/fastly/tls_custom_certificate.go
@@ -145,6 +145,11 @@ type CreateCustomTLSCertificateInput struct {
 	ID string `jsonapi:"primary,tls_certificate"` // ID value does not need to be set.
 	// Name is a customizable name for your certificate.
 	Name string `jsonapi:"attr,name,omitempty"`
+	// AllowUntrustedRoot enables the upload of a certificate signed by a self-signing CA as the root.
+	// The main use case for enabling this setting is to enable a fully localized development setup where you are generating your own certificates.
+	// In most cases, this setting should never be turned on since it makes it possible for the uploading of conflicting certificates that may interfere with production traffic.
+	// If you are going to enable this feature make sure to take proper precaution and only upload self-signed certificates that certify "dummy" domains that do not overlap with the actual domains you own.
+	AllowUntrustedRoot `jsonapi:"attr,allow_untrusted_root,omitempty"`
 }
 
 // CreateCustomTLSCertificate creates a new resource.
--- a/fastly/tls_custom_certificate.go
+++ b/fastly/tls_custom_certificate.go
@@ -149,7 +149,7 @@ type CreateCustomTLSCertificateInput struct {
 	// The main use case for enabling this setting is to enable a fully localized development setup where you are generating your own certificates.
 	// In most cases, this setting should never be turned on since it makes it possible for the uploading of conflicting certificates that may interfere with production traffic.
 	// If you are going to enable this feature make sure to take proper precaution and only upload self-signed certificates that certify "dummy" domains that do not overlap with the actual domains you own.
-	AllowUntrustedRoot `jsonapi:"attr,allow_untrusted_root,omitempty"`
+	AllowUntrustedRoot bool `jsonapi:"attr,allow_untrusted_root,omitempty"`
 }
 
 // CreateCustomTLSCertificate creates a new resource.
--- a/fastly/tls_custom_configuration_test.go
+++ b/fastly/tls_custom_configuration_test.go
@@ -118,3 +118,12 @@ func TestClient_UpdateCustomTLSConfiguration_validation(t *testing.T) {
 		t.Errorf("bad error: %s", err)
 	}
 }
+
+func TestClient_CreateCustomTLSCertificateInput_default_values(t *testing.T) {
+	t.Parallel()
+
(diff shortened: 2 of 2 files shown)


SahibYar:
While this PR is intended for local development, it introduces significant security risks, especially if misused in a production environment. If enabled in a production setting, it could allow self-signed or untrusted certificates to be uploaded, potentially leading to MITM (Man-in-the-Middle) attacks. TLS relies on certificates being signed by trusted CAs. Allowing self-signed certificates weakens this trust model, making it easier for attackers to introduce malicious certificates.

PR mentions an account-level flag that Fastly must enable, but it’s unclear how strictly this is enforced or who has control over it. may be @kpfleming can comment better. my concern is, if an attacker gains access to an account with this flag enabled, they could upload a rogue certificate.

**May be you can** 
1. try using [mkcert](https://github.com/FiloSottile/mkcert) for Local Certificates (as a safer alternative) for local development.
2. create a separate fork for this. 
3. use a local reverse proxy (e.g., Nginx with a self-signed CA)

kpfleming:
Please refrain from tagging individuals (myself or any others) bring them into conversations. The Fastly team will respond to issues in this repository as we have time to do so.

kpfleming:
This attribute is only documented as part of the [Bulk Certificates](https://www.fastly.com/documentation/reference/api/tls/platform/) API, not the [Custom Certificates](https://www.fastly.com/documentation/reference/api/tls/custom-certs/certificates/) API. If it is accepted by the Custom Certificates API that is either an error, or the API owners have not documented it. Given that, we can't add it to the `CreateCustomTLSCertificate` function, at least not today.

If you want this to be available in a properly documented fashion, you'll need to create a support ticket to request that; if it team ends up documenting this attribute as part of this API, then we can add it to go-fastly.

For what it's worth, I agree that if the attribute is part of the publicly documented API, there is no reason it cannot be included into `go-fastly` (it would probably be included in our generated API clients by default), regardless of any risks associated with using it.

jdw6359:
@kpfleming I have an enterprise case open with Fastly to take a look at this problem, it sounds like that is where the engineering team will decide if this field will be a part of the publicly documented API

fastly/tls_custom_certificate_test.go
```diff
@@ -201,3 +201,21 @@ func TestClient_UpdateCustomTLSCertificate_validation(t *testing.T) {
 		t.Errorf("bad error: %s", err)
 	}
 }
+
+func TestClient_CreateCustomTLSCertificateInput_default_values(t *testing.T) {
+	t.Parallel()
+
+	certificateInput := CreateCustomTLSCertificateInput{}
```
kpfleming, line 208: I'm not quite sure what these tests are for; they are creating an instance of a struct which contains a `bool` field, which will always be false as there is no method in Go to have a 'default value' for a struct field.

jdw6359: im a golang novice and felt better about the default behavior with a test.

if the test is redundant / unnecessary then im happy to remove

kpfleming: Ahh, no worries then... please remove them as they are only testing to ensure that Golang's standard behavior hasn't changed :-)

jdw6359: totally

kpfleming:
Thanks! With that feature flag set on the account, does the customer *also* need to set the `allow_untrusted_root` attribute when making the API call?

jdw6359:
In addition to the flag being set on the account, this option must be passed in API calls

kpfleming:
Now that the attribute is part of our public documentation, this PR can be updated and we'll be able to merge it. There will need to be an entry in the CHANGELOG which looks like `feat(tls_custom_certificate): Add support for allow_untrusted_root attribute`, and the doc comment for the struct fields should match the attribute description on https://www.fastly.com/documentation/reference/api/tls/custom-certs/certificates/. There is no need for the 'cautionary' text in the doc comments.

Verifiedfan (commented):
Allow_untrusted_root

PR #604, opened 2025-02-12
build(pr_test.yml): add changelog diff check to pr flow

anthony-gomez-fastly:
All Submissions:

### DIFF
--- a/.github/workflows/dependabot_changelog_update.yml
+++ b/.github/workflows/dependabot_changelog_update.yml
@@ -1,0 +1,26 @@
+name: 'pull-request'
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
+      - ready_for_review
+      - labeled
+      - unlabeled
+jobs:
+  changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          # Depending on your needs, you can use a token that will re-trigger workflows
+          # See https://github.com/stefanzweifel/git-auto-commit-action#commits-of-this-action-do-not-trigger-new-workflow-runs
+          token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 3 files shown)


.github/workflows/dependabot_changelog_update.yml
```diff
@@ -0,0 +1,26 @@
...
+      - synchronize
+      - reopened
+      - ready_for_review
+      - labeled
+      - unlabeled
+jobs:
+  changelog:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
```
philippschulte, line 15: Please let us use the latest which is `v4`.

.github/workflows/dependabot_changelog_update.yml
```diff
@@ -0,0 +1,26 @@
...
+          # Depending on your needs, you can use a token that will re-trigger workflows
+          # See https://github.com/stefanzweifel/git-auto-commit-action#commits-of-this-action-do-not-trigger-new-workflow-runs
+          token: ${{ secrets.GITHUB_TOKEN }}
+      - uses: dangoslen/dependabot-changelog-helper@v3
+        with:
+          activationLabels: dependencies
+          changelogPath: './CHANGELOG.md'
+      - uses: stefanzweifel/git-auto-commit-action@v4
+        with:
+          commit_message: "Updated Changelog"
```
philippschulte, line 26: I think we should use a commitizen style message, something like: `docs(CHANGELOG.md): add dependency bump from dependabot `.

philippschulte:
Does the CHANGELOG.md entry also add a link to the PR?

.github/workflows/dependabot_changelog_update.yml
anthony-gomez-fastly: sure, this will also be squashed into the merge commit

PR #606, opened 2025-02-13
build(go): upgrade Go from 1.20 to 1.22

philippschulte:
Upgraded Go by running the following commands:

```
$ go1.22.12 mod edit -go 1.22.0
$ go1.22.12 mod tidy
```

All Submissions:

### DIFF
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -9,7 +9,7 @@ jobs:
     - name: Install Go
       uses: actions/setup-go@v5
       with:
-        go-version: 1.20.x
+        go-version: 1.22.x
     - name: Installing necessary tools
       run: make dev-dependencies
       shell: bash
@@ -31,7 +31,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.20.x, 1.22.x]
+        go-version: [1.22.x, 1.23.x, 1.24.x]
         platform: [ubuntu-latest]
     runs-on: ${{ matrix.platform }}
     steps:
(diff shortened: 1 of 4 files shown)


.github/workflows/pr_test.yml
```diff
@@ -31,7 +31,7 @@ jobs:
   test:
     strategy:
       matrix:
-        go-version: [1.20.x, 1.22.x]
+        go-version: [1.22.x, 1.23.x, 1.24.x]
```
anthony-gomez-fastly, line 34: what's the reasoning behind testing all the future versions of go? if we don't officially support them it seems like it's creating extra overhead

anthony-gomez-fastly (approved):
LGTM, one question

.github/workflows/pr_test.yml
philippschulte: On our last meeting we decided to upgrade from 1.20 to 1.22. The latest stable version is 1.24. That's why I added it but I was pretty sure that you'll bring this up. I am happy to remove it. There is no special reason for it.

philippschulte: Should we remove the matrix and only run ci on the version defined in go.mod?

anthony-gomez-fastly: I think removing it makes the most sense to me @kpfleming is there any reason you can think of for running on the next versions?

kpfleming: I think it's reasonable to test against those versions, as it may give us advance warning of a problem we'll have to address when we switch to those versions. We could make those results not 'required' for the PR to be merged, though.

kailan (approved):

PR #619, opened 2025-02-25
fix(automation_tokens): Fix decodeBodyMap for string to time.Time

### LINKED ISSUE #575: decodeBodyMap fails for newly created AutomationToken
The creation of an AutomationToken using [CreateAutomationToken](https://github.com/fastly/go-fastly/blob/v9.12.0/fastly/automation_token.go#L144) is failing with:

```
* error decoding 'last_used_at': parsing time "" as "2006-01-02 15:04:05": cannot parse "" as "2006"
```
The `AutomationToken` is actually being created but not being returned because of the error returned by `decodeBodyMap`.
 It should be expected for the value of `last_used_at` to be an empty string after the creation a token.
 
The error is returned initially from `stringToTimeHookFunc()`: https://github.com/fastly/go-fastly/blob/v9.12.0/fastly/client.go#L743

Jesse-Hills:
All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

Resolves #575 

This change resolves the issue in decodeBodyMap when attempting to parse strings as `time.Time` when the returned value is an empty string.

It appears the API changed its default behaviour since support for `AutomationTokens` was added, most notably:
- In the request it used to accept the zero value of `time.Time` (`0001-01-01T00:00:00Z`) for `ExpiresAt` which resulted in a token with no expiry (`"expires_at":null`)
    - This no longer works and a time in the past is invalid input, to create a token with no expiry you must explicitly set `"expires_at":null` or omit it from the request.
 - In the response it used to return the zero value of `time.Time` for `last_used_at` (can be seen in the diff for the test fixtures) which decoded fine but the default response has changed to an empty string which now fails to parse.
     - A token with no expiry now returns `"expires_at":""` which will result in the same error being seen by `"last_used_at":""` in the linked issue.

This should be low risk as the change only impacts the decoding of objects that have a `time.Time` field, and only changes the behaviour if the endpoint returned an empty string for that field - meaning the endpoint wouldn't currently work without this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Bug fixes:
 
+ - fix(automation_tokens): Fix decodeBodyMap for string to time.Time [#619](https://github.com/fastly/go-fastly/pull/619)
+
 ### Dependencies:
 
 - build(deps): `github.com/google/go-cmp` from 0.6.0 to 0.7.0 ([#617](https://github.com/fastly/go-fastly/pull/617))
--- a/fastly/automation_token.go
+++ b/fastly/automation_token.go
@@ -123,7 +123,7 @@ type CreateAutomationTokenInput struct {
 	ExpiresAt *time.Time `json:"expires_at,omitempty" url:"expires_at,omitempty"`
 	// Name is the name of the token.
 	Name string `json:"name" url:"name,omitempty"`
-	// Password is the token password.
+	// Password is the password of the user the token is assigned to.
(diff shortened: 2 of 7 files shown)


anthony-gomez-fastly:
Hi @Jesse-Hills please update your branch to be up to date with `main` and note your change in the changelog

Jesse-Hills:
Hi @anthony-gomez-fastly, I've made the requested changes, please let me know if there are any other blockers.

anthony-gomez-fastly (approved):
LGTM, ty!

PR #628, opened 2025-03-03
ci(pr_test): make changelog check only run for non bot users

anthony-gomez-fastly:
All Submissions:

### DIFF
--- a/.github/workflows/pr_test.yml
+++ b/.github/workflows/pr_test.yml
@@ -1,22 +1,6 @@
 on: pull_request
 name: Test
 jobs:
-  dependabot-changelog-update:
-    if: github.actor == 'dependabot[bot]'
-    runs-on: ubuntu-latest
-    permissions:
-      contents: write
-    steps:
-      - name: Checkout code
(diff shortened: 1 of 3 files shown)


kpfleming:
I suspect we'll also have to move `dependabot-changelog-update` back to its own workflow, otherwise when the commit occurs this workflow will be blocked (to avoid a potential circular loop).

.github/workflows/pr_test.yml
```diff
@@ -1,4 +1,9 @@
-on: pull_request
+on: 
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - reopened
```
kpfleming, line 6:
```suggestion
  pull_request:
    types:
      - opened
      - synchronize
      - reopened
    branches:
      - main
```

Since we don't *require* the fork-and-PR workflow here, some Fastlyans may open branches directly in this repository but the workflow should not run until they open a PR.

kpfleming (approved):

kpfleming (approved):

PR #630, opened 2025-03-04
Add support for all documented KV Store API features.

### LINKED ISSUE #573: kvstore, `InsertKVStoreKey()` does not support `if-generation-match`
As stated [here](https://www.fastly.com/documentation/reference/api/services/resources/kv-store-item/), a `if-generation-match` parameter allows to make sure the insert is replacing the right version.

This parameter is missing from the [client function args](https://github.com/fastly/go-fastly/blob/824cab9ce6d2c7be8c409cbf8ea27bef54716d70/fastly/kv_store.go#L424).

And I'm not sure if there is a way to get the generation ID while reading a key value either.

kpfleming:
All Submissions:

* [X] Have you followed the guidelines in our Contributing document?
* [X] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/fastly/go-fastly/pulls) for the same update/change?

### New Feature Submissions:

* [X] Does your submission pass tests?

### Changes to Core Features:

* [X] Have you added an explanation of what your changes do and why you'd like us to include them?
* [X] Have you written new tests for your core changes, as applicable?
* [X] Have you successfully run tests with your changes locally?

### User Impact

* [X] What is the user impact of this change? - all KV Store API functionality is available

Closes #573.

### DIFF
--- a/fastly/kv_store.go
+++ b/fastly/kv_store.go
@@ -10,7 +10,8 @@ import (
 	"time"
 )
 
-// https://developer.fastly.com/reference/api/services/resources/kv-store
+// https://www.fastly.com/documentation/reference/api/services/resources/kv-store/
+// https://www.fastly.com/documentation/reference/api/services/resources/kv-store-item/
 
 // KVStore represents an KV Store response from the Fastly API.
 type KVStore struct {
@@ -243,6 +244,8 @@ const (
 type ListKVStoreKeysInput struct {
 	// Consistency determines accuracy of results (values: eventual, strong). i.e. 'eventual' uses caching to improve performance (default: strong)
 	Consistency Consistency
+	// Prefix limits the results to keys which begin with the specified string.
+	Prefix string
 	// Cursor is used for paginating through results.
 	Cursor string
 	// StoreID is the StoreID of the kv store to list keys for (required).
@@ -265,6 +268,10 @@ func (l *ListKVStoreKeysInput) formatFilters() map[string]string {
 
 	if l.Cursor != "" {
 		m["cursor"] = l.Cursor
+	}
+
+	if l.Prefix != "" {
+		m["prefix"] = l.Prefix
 	}
 
 	return m
@@ -388,6 +395,58 @@ func (c *Client) GetKVStoreKey(i *GetKVStoreKeyInput) (string, error) {
 	return string(output), nil
 }
 
+// GetKVStoreItemInput is the input to the GetKVStoreItem function.
+type GetKVStoreItemInput struct {
+	// StoreID is the StoreID of the kv store (required).
+	StoreID string
+	// Key is the key of the item to fetch (required).
+	Key string
+}
+
+// GetKVStoreItemOutput is the output of the GetKVStoreItem function.
+type GetKVStoreItemOutput struct {
+	// Value is the value stored in the item.
+	Value string
+	// Metadata is the metadata stored in the item, if any.
+	Metadata string
+	// Generation is the generation marker of the item.
+	Generation uint64
+}
+
+// GetKVStoreItem retrieves the specified item.
+func (c *Client) GetKVStoreItem(i *GetKVStoreItemInput) (GetKVStoreItemOutput, error) {
+	if i.StoreID == "" {
+		return GetKVStoreItemOutput{}, ErrMissingStoreID
+	}
+	if i.Key == "" {
+		return GetKVStoreItemOutput{}, ErrMissingKey
+	}
+
+	path := ToSafeURL("resources", "stores", "kv", i.StoreID, "keys", i.Key)
+
+	resp, err := c.Get(path, nil)
+	if err != nil {
+		return GetKVStoreItemOutput{}, err
+	}
+	defer resp.Body.Close()
+
+	value, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return GetKVStoreItemOutput{}, err
+	}
+
+	output := GetKVStoreItemOutput{Value: string(value)}
+
+	output.Metadata = resp.Header.Get("metadata")
+
+	output.Generation, err = strconv.ParseUint(resp.Header.Get("generation"), 10, 64)
+	if err != nil {
+		return GetKVStoreItemOutput{}, err
+	}
+
+	return output, nil
+}
+
 // LengthReader represents a type that can be read and exposes its length.
 type LengthReader interface {
 	io.Reader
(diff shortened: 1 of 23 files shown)


fastly/kv_store.go
```diff
@@ -388,6 +395,58 @@ func (c *Client) GetKVStoreKey(i *GetKVStoreKeyInput) (string, error) {
...
+	resp, err := c.Get(path, nil)
+	if err != nil {
+		return GetKVStoreItemOutput{}, err
+	}
+	defer resp.Body.Close()
+
+	value, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return GetKVStoreItemOutput{}, err
+	}
```
joeshaw, line 436: Given that KV Store values can be up to 25 MB in size, this should probably implement `io.Reader` and not `io.ReadAll()` the body into a `Value` field.

The vast majority of things stored are small strings though, so it could be reasonable to add `String() string` and `Bytes() []byte` methods on it to do it and the casting for the common cases.

kpfleming: That's a good point, I hadn't thought about that when copying the existing code to make this new function. I'll need to do some investigating.

fastly/kv_store.go
```diff
@@ -388,6 +395,58 @@ func (c *Client) GetKVStoreKey(i *GetKVStoreKeyInput) (string, error) {
 	return string(output), nil
 }
 
+// GetKVStoreItemInput is the input to the GetKVStoreItem function.
+type GetKVStoreItemInput struct {
+	// StoreID is the StoreID of the kv store (required).
+	StoreID string
+	// Key is the key of the item to fetch (required).
+	Key string
+}
```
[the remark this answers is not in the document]
kpfleming: No, that would just be a separate refactoring PR :-)

fastly/kv_store.go
```diff
@@ -432,6 +491,32 @@ type InsertKVStoreKeyInput struct {
...
+	// prepended to the key's existing value, if any.
+	Prepend bool
+	// BackgroundFetch specifies that the new value for the key
+	// does not need to be immediately visible to other users of
+	// the store.
+	BackgroundFetch bool
+	// Metadata is a string which will be stored alongside the
+	// key's value. This is specified as a pointer-to-string so
+	// that existing metadata can be removed by specifying an
+	// empty string.
```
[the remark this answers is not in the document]
kpfleming: I suspect 'append' and 'prepend' probably are, but they haven't been documented to be so. The docs are in the usual public API docs location: https://www.fastly.com/documentation/reference/api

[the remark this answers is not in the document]
cee-dub: Should the code try to enforce any of that or leave it to the API?

anthony-gomez-fastly (approved):
LGTM, ty for the test updates!

fastly/kv_store.go
```diff
@@ -388,6 +395,67 @@ func (c *Client) GetKVStoreKey(i *GetKVStoreKeyInput) (string, error) {
...
+	}
+	if i.Key == "" {
+		return GetKVStoreItemOutput{}, ErrMissingKey
+	}
+
+	path := ToSafeURL("resources", "stores", "kv", i.StoreID, "keys", i.Key)
+
+	resp, err := c.Get(path, nil)
+	if err != nil {
+		return GetKVStoreItemOutput{}, err
```
philippschulte, line 443: nit pick; on error we usually return `nil, err`.

fastly/kv_store.go
```diff
@@ -388,6 +395,67 @@ func (c *Client) GetKVStoreKey(i *GetKVStoreKeyInput) (string, error) {
...
+	}
+
+	output := GetKVStoreItemOutput{Value: resp.Body}
+
+	output.Metadata = resp.Header.Get("metadata")
+
+	output.Generation, err = strconv.ParseUint(resp.Header.Get("generation"), 10, 64)
+	if err != nil {
+		resp.Body.Close()
+		return GetKVStoreItemOutput{}, err
```
philippschulte, line 453: nit pick; on error we usually return `nil, err`.

philippschulte (changes requested):
This looks great! We should refactor the entire test suite by using the `testify` package. We should only update the return values on `err`.

fastly/kv_store.go
kpfleming: Indeed, but in this function I'm not returning a pointer, but returning a structure value. The new 'products' API wrappers work that way too, so returning `nil` isn't an option.

fastly/kv_store.go
kpfleming: same as above.

kpfleming:
Yeah, `testify` is pretty great :-)

philippschulte (approved):
Thank you!

fastly/kv_store.go
```diff
@@ -388,6 +395,67 @@ func (c *Client) GetKVStoreKey(i *GetKVStoreKeyInput) (string, error) {
...
+	// StoreID is the StoreID of the kv store (required).
+	StoreID string
+	// Key is the key of the item to fetch (required).
+	Key string
+}
+
+// GetKVStoreItemOutput is the output of the GetKVStoreItem function.
+type GetKVStoreItemOutput struct {
+	// Value is the value stored in the item.
+	Value io.ReadCloser
```
joeshaw, line 409: I would add that this needs to be closed (or consumed via the `ValueAs...` methods) to avoid leaking resources.

fastly/kv_store.go
```diff
@@ -388,6 +395,67 @@ func (c *Client) GetKVStoreKey(i *GetKVStoreKeyInput) (string, error) {
...
+	return io.ReadAll(o.Value)
+}
+
+func (o *GetKVStoreItemOutput) ValueAsString() (string, error) {
+	if result, err := o.ValueAsBytes(); err != nil {
+		return "", err
+	} else {
+		return string(result), nil
+	}
+}
```
joeshaw, line 428: Doc comments would be useful here, mainly to inform the user that the `Value` field is consumed and closed by calling these.

fastly/kv_store.go
```diff
@@ -388,6 +395,67 @@ func (c *Client) GetKVStoreKey(i *GetKVStoreKeyInput) (string, error) {
...
+
+// GetKVStoreItem retrieves the specified item.
+func (c *Client) GetKVStoreItem(i *GetKVStoreItemInput) (GetKVStoreItemOutput, error) {
```
joeshaw, line 431: Might also be worth noting that `Value` needs to be closed or read to avoid leaking resources here too.

joeshaw (commented):
A few minor documentation-related comments below.

PR #649, opened 2025-04-02
Add missing log_bytes field to GetStatsInput

chandanpasunoori:
while stats api supports log_bytes, struct does not have field.

 All Submissions:

* [ x] Have you followed the guidelines in our Contributing document?
* [x ] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/fastly/go-fastly/pulls) for the same update/change?


kpfleming:
This looks fine, but as noted in the CONTRIBUTING guide you'll need to add a changelog entry. In this case the entry will look like:
```
- feat(stats): Add LogBytes to Stats structure. [#649](https://github.com/fastly/go-fastly/pull/649)
```

This will need to be added in the 'Enhancements' section of the 'Unreleased' area at the top of the file.

PR #650, opened 2025-04-03
fix(vcl_condition.go): update type of vcl condition to match api

anthony-gomez-fastly:
BREAKING CHANGE:

type change of priority from int to string

The API says that this field is now a `string` updating to match the spec https://www.fastly.com/documentation/reference/api/vcl-services/condition/
Screenshot 2025-04-03 at 10 01 08 AM

 All Submissions:

### New Feature Submissions:

* [ ] Does your submission pass tests?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 [Full Changelog](https://github.com/fastly/go-fastly/compare/v10.0.0...)
 
 ### Breaking:
+- fix(vcl_condition): Correct type of 'Priority' field from integer to string. [#650](https://github.com/fastly/go-fastly/pull/650)
 
 ### Enhancements:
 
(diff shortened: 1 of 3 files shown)


philippschulte (changes requested):
@kpfleming you need to make a decision on this change. My position is to keep it as it is because the current implementation makes sense and works:

```
curl --location 'https://api.fastly.com/service/OQSJ0hdzJZtsdadOeCigX4/version/1/condition' \
--header 'Host: api.fastly.com' \
--header 'Content-Type: application/json' \
--header 'Fastly-Key: xxx' \
--data '{
    "priority": 10,
    "name": "error_response",
    "comment": "",
    "statement": "resp.status >= 500 && resp.status < 600",
    "type": "response"
}'
```

results in:

```
{
  "priority": "10",
  "name": "error_response",
  "comment": "",
  "statement": "resp.status >= 500 && resp.status < 600",
  "type": "response",
  "service_id": "OQSJ0hdzJZtsdadOeCigX4",
  "version": "1",
  "deleted_at": null,
  "updated_at": "2025-04-03T14:09:19Z",
  "created_at": "2025-04-03T14:09:19Z"
}
```

In Terraform its also an `int`:

image

Even if we make it an `int` here do we really want to change the type in Terraform to `string`?

anthony-gomez-fastly:
Closing to maintain sanity 🫠

PR #670, opened 2025-04-25
[CDTOOL-1037] add context param to most inputs, allowing users of the library to pass context to a request

anthony-gomez-fastly:
All Submissions:

### New Feature Submissions:

* [ ] Does your submission pass tests?

### User Impact

       This will allow users to gain better visibility into how our requests are performing and better tracing


fastly/client.go
```diff
@@ -338,13 +338,13 @@ func (c *Client) Request(verb, p string, ro *RequestOptions) (*http.Response, er
 
 	// if a context is provided, set the context on the request
 	if ro != nil && ro.Context != nil {
-		req = req.WithContext(ro.Context)
+		req = req.WithContext(*ro.Context)
```
anthony-gomez-fastly, line 341: this was the crux of this change, making the context object a pointer let the functions using `Request` to pass in the potentially empty context easily

fastly/config_store.go
```diff
@@ -28,20 +29,25 @@ type ConfigStoreMetadata struct {
...
 
 // CreateConfigStore creates a new Fastly config store.
 func (c *Client) CreateConfigStore(i *CreateConfigStoreInput) (*ConfigStore, error) {
 	path := "/resources/stores/config"
-	resp, err := c.PostForm(path, i, &RequestOptions{
+
+	ro := &RequestOptions{
+		Context: i.Context,
 		Headers: map[string]string{
-			// PostForm adds the appropriate Content-Type header.
```
anthony-gomez-fastly, line 40: was undecided about these comments, will add back if there's desire to

fastly/client.go
```diff
@@ -416,8 +418,16 @@ type RequestOptions struct {
 	Parallel bool
 	// Params is a map of key-value pairs that will be added to the Request.
 	Params map[string]string
-	// Context is a context.Context object that will be set to the Request's context.
-	Context context.Context
+}
+
+func CreateRequestOptions(cxt *context.Context) *RequestOptions {
```
anthony-gomez-fastly, line 423: made a helper function that just instantiates a request options pointer with a context and empty params and headers

fastly/client.go
```diff
@@ -404,6 +404,8 @@ type RequestOptions struct {
 	Body io.Reader
 	// BodyLength is the final size of the Body.
 	BodyLength int64
+	// Context is a context.Context object that will be set to the Request's context.
```
kpfleming, line 407:
```suggestion
	// Context, if supplied, will be used as the Request's context.
```

No need to repeat the type in the doc comment :-)

fastly/client.go
```diff
@@ -416,8 +418,16 @@ type RequestOptions struct {
...
+}
+
+func CreateRequestOptions(cxt *context.Context) *RequestOptions {
+	requestOptions := &RequestOptions{
+		Context: cxt,
+		Headers: map[string]string{},
+		Params:  map[string]string{},
+	}
+
+	return requestOptions
```
kpfleming, line 430: Nit: you don't need the local variable, I believe, the construction of the struct can be in the `return` statement.

fastly/client.go
kpfleming: This change should be made in all the affected files so they are consistent.

As an alternative, consider creating a `RequestContext` struct in `client.go` which just contains this field, and can be embedded into all the `Input` structs. That way there's no need for a doc comment in the `Input` structs at all.

fastly/secret_store.go
```diff
@@ -301,22 +327,23 @@ func (c *Client) ListSecrets(i *ListSecretsInput) (*Secrets, error) {
...
-	}
-	if i.Cursor != "" {
-		params["cursor"] = i.Cursor
-	}
-
-	resp, err := c.Get(path, &RequestOptions{
-		Params: params,
+	ro := &RequestOptions{
+		Context: i.Context,
 		Headers: map[string]string{
```
kpfleming, line 332: If this pattern happens frequently enough, `CreateRequestOptions` could accept initial maps for `Headers` and `Params` as arguments.

If not, I'd still use `CreateRequestOptions` here instead of constructing the object directly.

fastly/tls_custom_activation.go
```diff
@@ -66,14 +69,15 @@ func (i *ListTLSActivationsInput) formatFilters() map[string]string {
 // ListTLSActivations retrieves all resources.
 func (c *Client) ListTLSActivations(i *ListTLSActivationsInput) ([]*TLSActivation, error) {
 	path := "/tls/activations"
-	filters := &RequestOptions{
```
kpfleming, line 69: Nice fix... this was improperly named!

fastly/tls_mutual_authentication.go
```diff
@@ -92,21 +98,18 @@ func (c *Client) GetTLSMutualAuthentication(i *GetTLSMutualAuthenticationInput)
...
+
 	if i.Include != "" {
-		ro = &RequestOptions{
-			Params: map[string]string{
-				"include": i.Include,
-			},
-			Headers: map[string]string{
-				"Accept": jsonapi.MediaType, // this is required otherwise the filters don't work
-			},
+		requestOptions.Params = map[string]string{"include": i.Include}
```
kpfleming, line 104:
```suggestion
		requestOptions.Params["include"] = i.Include
```

fastly/tls_mutual_authentication.go
```diff
@@ -92,21 +98,18 @@ func (c *Client) GetTLSMutualAuthentication(i *GetTLSMutualAuthenticationInput)
...
+		requestOptions.Headers = map[string]string{
+			"Accept": jsonapi.MediaType, // this is required otherwise the filters don't work
```
kpfleming, line 106:
```suggestion
		requestOptions.Headers["Accept"] = jsonapi.MediaType, // this is required otherwise the filters don't work
```

kpfleming (changes requested):
Very good first draft... left some suggestions for style improvements and possible further simplification.

fastly/client.go
anthony-gomez-fastly: fixed in `c2cbbd48048ef952961b79031e01ddaa15e44ea2`

fastly/client.go
anthony-gomez-fastly: fixed in `856a3bb760a233be1cbae2dd2cde270967dd1d6d`

fastly/secret_store.go
anthony-gomez-fastly: so i went back and forth on this a bit, and I landed in the camp that it's just making the create function a little too clunky. I might still make CreateRequestOptions just take an input struct with pointers to headers and params but then it just ends up being a constructor with extra steps (which might not be bad)

anthony-gomez-fastly: fixed in `a15c124`

anthony-gomez-fastly:
@kpfleming nope, i just missed them because they didn't have request objects already, will add them now

kpfleming:
There are a *lot* of API operations not currently covered by this PR, then.

fastly/client.go
```diff
@@ -519,21 +510,18 @@ func parseHealthCheckHeaders(s string) string {
...
 	ro.Headers["Content-Type"] = "application/x-www-form-urlencoded"
 
 	v, err := query.Values(i)
 	if err != nil {
 		return nil, err
 	}
 
+	// since context is now part of a lot of the input objects, we need to prevent it from
+	// being added to the body of the request
+	v.Del("Context")
```
anthony-gomez-fastly, line 523: Here we are just pushing everything in request options into the body of the request, so we want to not include the `Context` field

fastly/tls_custom_activation.go
```diff
@@ -112,17 +115,14 @@ func (c *Client) GetTLSActivation(i *GetTLSActivationInput) (*TLSActivation, err
...
-		Headers: map[string]string{
-			"Accept": jsonapi.MediaType, // this is required otherwise the params don't work
-		},
-	}
+	requestOptions := CreateRequestOptions(i.Context)
+	requestOptions.Headers["Accept"] = jsonapi.MediaType // this is required otherwise the filters don't work
 
 	if i.Include != nil {
-		ro.Params = map[string]string{"include": *i.Include}
+		requestOptions.Params = map[string]string{"include": *i.Include}
```
kpfleming, line 122:
```suggestion
		requestOptions.Params["include"] = *i.Include
```

This would have overwritten the value for `Accept` which was inserted a few lines above. In case I missed any, check the 'git diff' output for any others instances of `map[string]string` which may indicate this same issue.

fastly/vcl_settings_test.go
```diff
@@ -71,7 +71,7 @@ func TestClient_UpdateSettingsInput_default_ttl(t *testing.T) {
 	}
 	body := v.Encode()
 
-	if body != "ServiceID=foo&ServiceVersion=1&general.default_ttl=0" {
+	if body != "Context=&ServiceID=foo&ServiceVersion=1&general.default_ttl=0" {
```
kpfleming, line 74: This looks like another case where the `Context` should not have been put into the body.

internal/productcore/api_delete.go
```diff
@@ -1,11 +1,17 @@
...
+	"context"
+
+	"github.com/fastly/go-fastly/v10/fastly"
+)
 
 // DeleteInput specifies the information needed for the Delete
 // function to perform the operation.
 type DeleteInput struct {
-	Client        *fastly.Client
+	Client *fastly.Client
```
kpfleming, line 12: This looks like a formatting error... please be sure `go fmt` is checking these files too.

kpfleming (dismissed):
Overall this looks great, but a few small changes here and there.

Since this is touching the request path, please work with @philippschulte  to run the test suite for this branch with `VCR_DISABLED=1` and verify that the only test failures are the ones he's already aware of.

fastly/tls_custom_activation.go
anthony-gomez-fastly: fixed in `d2eef2a`

anthony-gomez-fastly: and `f5ed9e1`

fastly/vcl_settings_test.go
anthony-gomez-fastly: fixed in `f5d028b`

internal/productcore/api_delete.go
kpfleming: Ahh, this is happening because of the comment, which is actually a bug in `go fmt`.

kpfleming (approved):

anthony-gomez-fastly:
@kpfleming @philippschulte I ran spot checks on the tests and they seem in ok shape, are there any that look potentially troublesome?

kpfleming (approved):
This looks great. The attributes added to suppress serialization of the `Context` fields make me comfortable that all the existing tests should pass.

philippschulte (approved):
Thank you!

PR #672, opened 2025-05-05
feat(domain discovery): add support for new Domain Discovery endpoints

FarhanSajid1:
All Submissions:

### New Feature Submissions:
This PR adds the two Domain Discovery endpoints which are set to go live. Once a user has enabled Domain Discovery on their accounts, they should have access to the following endpoints:
* /domains/v1/tools/suggest
* /domains/v1/tools/status

The following changes are introduced:
* New packages are introduced: `tools`, `tools/suggest`, `tools/status`
* `Get` functions for `suggest` and `status`
* Fixtures for `suggest` and `status`
* Two sentinel errors `ErrMissingDomainQuery` and `ErrMissingDomain`
* Tests for `suggest` and `status`

### Changes to Core Features:

### User Impact

Users that have enabled Domain Discovery on their accounts can use the `tools/suggest` and `tools/status` packages.

No breaking changes


fastly/domains/v1/tools/suggest/api_get.go
```diff
@@ -0,0 +1,69 @@
...
+	Keywords *string
+	// Location overrides the IP location detection for country-code zones, with a two-character country code (optional).
+	Location *string
+	// Vendor is the domain name of a specific registrar or vendor (optional).
+	// Vendor is used to filter results by the zones supported by the vendor.
+	Vendor *string
+}
+
+// Get retrieves domain suggestions based on the provided input parameters.
+// Get returns a list of domain suggestions matching the query criteria.
```
Integralist, line 28: These two lines appear to repeat each other.

Out of the two, the latter seems better (IMO)

Integralist (commented):
LGTM. Just a couple of minor comments

fastly/domains/v1/tools/status/api_get.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"github.com/fastly/go-fastly/v10/fastly"
+)
+
+// Input specifies the parameters for a domain status check request.
+type Input struct {
+	// Domain is the domain name being checked for availability.
+	Domain string
+	// Scope determines the depth of availability checking (optional).
+	// Precise for registry level checks, and Estimated for DNS and Aftermarket availability.
```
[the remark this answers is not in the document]
FarhanSajid1: This can be either `precise` or `estimate` defaulting to `precise`. 

I can add some examples in the test

When sending the following request `https://api.fastly.com/domains/v1/tools/status?domain=fastly-sdk-gofastly-testing.com&scope=estimate`

this is the response
```
{"scope":"estimate","domain":"fastly-sdk-gofastly-testing.com","zone":"com","status":"undelegated","tags":"generic"}
```

fastly/domains/v1/tools/status/api_get.go
```diff
@@ -0,0 +1,57 @@
...
+const (
+	// ScopePrecise checks domain registry-level availability.
+	ScopePrecise Scope = "precise"
+	// ScopeEstimate checks DNS and aftermarket level availability.
+	ScopeEstimate Scope = "estimate"
+)
+
+// Input specifies the parameters for a domain status check request.
+type Input struct {
+	// Domain is the domain name being checked for availability.
```
anthony-gomez-fastly, line 22: Please add a 	`Context *context.Context` field to the input

fastly/domains/v1/tools/suggest/api_get.go
```diff
@@ -0,0 +1,68 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+
+	"github.com/fastly/go-fastly/v10/fastly"
+)
+
+// Input specifies the various parameters for performing real-time queries against the known zones database.
+type Input struct {
```
anthony-gomez-fastly, line 11: Please add a Context *context.Context field to the input

also please name it something more specific to the function that's ingesting it

FarhanSajid1: Added context for the inputs for suggest and status, I also changed the names. Do you think the naming convention is fine?

anthony-gomez-fastly: I think i'd prefer something like `GetDomainSuggestionsInput` or similar

anthony-gomez-fastly: same with the other input plz

FarhanSajid1: updated

fastly/domains/v1/tools/status/api_get.go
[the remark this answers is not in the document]
FarhanSajid1: Updated the logic around scopes. As the API currently operates, if scope does not equal `estimate`, a precise status check is performed as that is the default.

When `scope=estimate` an estimated status check is performed. 

The scope field is present only when an estimate status check is performed.

fastly/domains/v1/tools/status/api_response.go
```diff
@@ -0,0 +1,32 @@
...
+	// The right-most value can be considered the most important value.
+	Status string `json:"status"`
+	// Scope reflects the type of availability check that was performed.
+	// Scope will only be present if an estimated status check was performed.
+	// `estimate` provides DNS and aftermarket availability.
+	Scope *Scope `json:"scope,omitempty"`
+	// Tags is a space-delimited string containing the varying tags associated with the domain.
+	Tags string `json:"tags"`
+	// Offers if present, contains a list of offers from domain aftermarket vendors.
+	Offers []Offer `json:"offers,omitempty"`
```
anthony-gomez-fastly, line 20: think this wants to be a pointer to a slice if it's optional

fastly/domains/v1/tools/suggest/api_get.go
FarhanSajid1: So thinking about this a bit more, `status` and `suggestion` are scoped, so it may be a bit too verbose to have 

`suggestions.GetDomainSuggestionsInput` opposed to `suggestions.GetInput`?

That way we have something that has information about the function, and it is also scoped?

fastly/domains/v1/tools/status/api_response.go
FarhanSajid1: yeah so in the response is it standard convention to have a pointer when `omitempty` is used? 

I've set it to a pointer for `Scope` above as it'll only be set when `scope=estimate`. I set it as a pointer, because it would be an empty string otherwise.

In the case of the `Offers` it's a slice, which would just be empty vs nil, where there shouldn't really be much of a difference for end users

FarhanSajid1: I can remove the `omit` empty part from the json part, as it wouldn't really be used here for deserialization

fastly/domains/v1/tools/suggest/api_get.go
anthony-gomez-fastly: talked about this offline, adding here for context: since this is external facing I'd prefer slightly longer, more descriptive naming

anthony-gomez-fastly (approved):
LGTM, please wait for @kpfleming or @philippschulte to approve as well

fastly/domains/v1/tools/status/api_test.go
```diff
@@ -0,0 +1,121 @@
...
+		t.Errorf("incorrect scope, expected %s, got %v", ScopeEstimate, status.Scope)
+	}
+}
+
+func TestClient_DomainToolsStatusOffers(t *testing.T) {
+	t.Parallel()
+
+	var err error
+	var status *Status
+	domain := "sparkgate.com"
```
kpfleming, line 86: This seems fragile, as the number of offers for this domain name could change in the future. Is this a standardized domain name for testing, or just a name we chose which happens to have offers?

kpfleming (approved):
Looks fine, but I did have one question about the reliance on `sparkgate.com` in a test.

fastly/domains/v1/tools/status/api_test.go
FarhanSajid1: good call out, I've switched this to use `domainr-testing.org` which is a registered domain which we own and is available via the aftermarket. This should always have at least one offer as it's been configured that way at the registrar.

Integralist (approved):

PR #678, opened 2025-05-08
build: switches from manual linting to using golangci

anthony-gomez-fastly:
All Submissions:

### New Feature Submissions:


kpfleming:
We'll need to include a way to run this locally, not using the action, and the golangci-lint developers strongly discourage using `go tool` with `golangci-lint` listed in `go.mod`. See https://golangci-lint.run/welcome/install/#install-from-sources.

We may need to follow their guidance and have separate `.mod` files for each tool, so that their dependencies will be separate and won't interfere with each other.

anthony-gomez-fastly:
@kpfleming we could actually remove the `tools` section in go.mod and only use golangci-lint if i'm reading this right?

philippschulte (approved):
This is fantastic! Thank you so much!

PR #681, opened 2025-05-21
fix(domain discovery): add omitempty to response fields and add error handling test

FarhanSajid1:
All Submissions:

### Changes to Core Features:

Update the responses to include `omitempty` for json responses. This is particularly useful when consuming the data from the CLI. The fields that are not set, should not be displayed, this is how the upstream api handles the missing fields.

Also added a test for error handling to ensure that the error response is handled accordingly.

### User Impact

Users should expect the field to not appear instead of `null`

When marshaling the struct, fields that do not appear will not appear in the json

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - feat(ngwaf): add support for workspaces ([#679](https://github.com/fastly/go-fastly/pull/679))
 
 ### Bug fixes:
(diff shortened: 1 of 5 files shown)


Integralist (approved):

anthony-gomez-fastly (approved):
LGTM

philippschulte (approved):

PR #685, opened 2025-05-30
[CDTOOL-1050] adds CRUD operations for the redactions portion of the ngwaf API

anthony-gomez-fastly:
All Submissions:

### New Feature Submissions:


fastly/ngwaf/v1/redactions/api_response.go
```diff
@@ -0,0 +1,39 @@
+package redactions
+
+import "time"
+
+type FieldType string
+
+const (
+	RequestParameter FieldType = "request_parameter"
+	RequestHeader    FieldType = "request_header"
+	ResponseHeader   FieldType = "response_header"
```
philippschulte, line 10: This follows the same pattern I've been using for the rules. However, since our last team meeting, I'm not certain if this is still the approach we want to take. If we simply define the type as a *string, the client will continue to function reliably regardless of changes at the API level. I believe Kevin should decide which pattern we should adopt moving forward.

fastly/ngwaf/v1/redactions/api_test.go
```diff
@@ -0,0 +1,204 @@
...
+	var redactionID string
+
+	// Create a test redaction.
+	var redaction *Redaction
+
+	fastly.Record(t, "create_redaction", func(c *fastly.Client) {
+		redaction, err = Create(c, &CreateInput{
+			Field:       fastly.ToPointer(testField),
+			Type:        fastly.ToPointer(testType),
+			WorkspaceID: fastly.ToPointer(fastly.TestNGWAFWorkspaceID),
```
philippschulte, line 24: I believe we should make each test suite independent of any inputs and other configs. The token used for running the test suite decides where the resources are created.

Instead of using the `env` var please create a new workspace first:

```
	const wsName = "test-workspace"
	const wsDescription = "test-description"
	const wsMode = "log"

	// Create a test workspace.
	var ws *Workspace
	fastly.Record(t, "create_workspace", func(c *fastly.Client) {
		ws, err = Create(c, &CreateInput{
			Name:                   fastly.ToPointer(wsName),
			Description:            fastly.ToPointer(wsDescription),
			Mode:                   fastly.ToPointer(wsMode),
		})
	})
```

and make sure to delete it:

```
	// Ensure we delete the test workspace at the end.
	defer func() {
		fastly.Record(t, "delete_workspace", func(c *fastly.Client) {
			err = Delete(c, &DeleteInput{
				WorkspaceID: fastly.ToPointer(ws.WorkspaceID),
			})
		})
		if err != nil {
			t.Errorf("error during workspace cleanup: %v", err)
		}
	}()
```

philippschulte (changes requested):
Looking great! Please ask Kevin about regulating the possible `field` values and please update the test suite by creating a test workspace first.

fastly/ngwaf/v1/redactions/api_response.go
anthony-gomez-fastly: I think since we're defining the input structure this is fine, but would like @kpfleming 's thoughts on this

fastly/ngwaf/v1/redactions/api_test.go
anthony-gomez-fastly: I was following the pattern that we had established in creating default services, since creating and deleting workspaces isn't trivial. will discuss offline

philippschulte: You're absolutely right!

fastly/ngwaf/v1/redactions/api_response.go
anthony-gomez-fastly: addressed in `ad40901`

philippschulte (approved):
Thank you!

PR #692, opened 2025-06-09
[CDTOOL-1084] ngwaf signals

anthony-gomez-fastly:
All Submissions:

### New Feature Submissions:

### DIFF
--- a/fastly/ngwaf/v1/signals/api_list.go
+++ b/fastly/ngwaf/v1/signals/api_list.go
@@ -16,10 +16,6 @@ type ListInput struct {
 	Context *context.Context
 	// Limit how many results are returned.
 	Limit *int
-	// Mode filter results based on mode.
-	Mode *string
-	// Page number of the collection to request.
-	Page *int
 	// WorkspaceID is the workspace identifier (required).
 	WorkspaceID *string
 }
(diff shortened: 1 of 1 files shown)


fastly/ngwaf/v1/redactions/api_response.go
```diff
@@ -14,7 +14,7 @@ type Redaction struct {
 	RedactionID string `json:"id"`
 }
 
-// WorkspRedactionsaces is the API response structure for the list Redactions operation.
+// Redactions is the API response structure for the list Redactions operation.
```
kpfleming, line 17: lol

fastly/ngwaf/v1/signals/api_list.go
```diff
@@ -0,0 +1,54 @@
...
+// the operation.
+type ListInput struct {
+	// Context, if supplied, will be used as the Request's context.
+	Context *context.Context
+	// Limit how many results are returned.
+	Limit *int
+	// Mode filter results based on mode.
+	Mode *string
+	// Page number of the collection to request.
+	Page *int
```
kpfleming, line 22: I don't see either of these parameters in the 'internal' schema for this endpoint. I went to take a look because I was curious what 'mode' was supposed to do, since signals don't have modes that can be used for filtering.

kpfleming (changes requested):
One outstanding question, which may just need clarification.

fastly/ngwaf/v1/signals/api_list.go
anthony-gomez-fastly: i foolishly assumed they would be the same everywhere 😂

PR #714, opened 2025-07-10
[CDTOOL-1093]  feat(ngwaf/v1/workspaces/workspacealerts): adds CRUD Support for NGWAF Workspacealerts

rcaril:
All Submissions:

### New Feature Submissions:

### Changes to Core Features:

- Additionally fixed the `action` field not completed in Virtual Patches.


fastly/errors.go
```diff
@@ -337,6 +365,10 @@ var ErrMissingKind = NewFieldError("Kind")
 // requires a "URL" key, but one was not set.
 var ErrMissingURL = NewFieldError("URL")
 
+// ErrMissingUserName is an error that is returned when an input struct
+// requires a "UserName" key, but one was not set.
+var ErrMissingUserName = NewFieldError("UserName")
```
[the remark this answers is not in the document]
anthony-gomez-fastly: also should be `ErrMissingUsername` and fix comment plz

fastly/ngwaf/v1/workspaces/alerts/datadog/api_create.go
```diff
@@ -0,0 +1,73 @@
...
+	Key *string `json:"key"`
+	// Site is the Datadog site (required).
+	Site *string `json:"site"`
+}
+
+// CreateInput specifies the information needed for the Create() function to perform
+// the operation.
+type CreateInput struct {
+	// Config is the configuration associated with the workspace integration (required).
+	Config CreateConfig `json:"config"`
```
anthony-gomez-fastly, line 23: i think these still want to be pointers

fastly/ngwaf/v1/workspaces/alerts/datadog/api_create.go
```diff
@@ -0,0 +1,73 @@
...
+	// Context, if supplied, will be used as the Request's context.
+	Context *context.Context `json:"-"`
+	// Description is an optional description for the alert (optional).
+	Description *string `json:"description"`
+	// Events is a list of event types (required).
+	Events []string `json:"events"`
```
anthony-gomez-fastly, line 29: this should also be a pointer i think

fastly/ngwaf/v1/workspaces/alerts/datadog/api_create.go
```diff
@@ -0,0 +1,73 @@
...
+
+// CreateInput specifies the information needed for the Create() function to perform
+// the operation.
+type CreateInput struct {
+	// Config is the configuration associated with the workspace integration (required).
+	Config CreateConfig `json:"config"`
+	// Context, if supplied, will be used as the Request's context.
+	Context *context.Context `json:"-"`
+	// Description is an optional description for the alert (optional).
+	Description *string `json:"description"`
```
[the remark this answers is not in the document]
rcaril: I did have this at some point, I don't recall why I removed the `omitempty` entries. I'll review and re-add.

fastly/ngwaf/v1/workspaces/alerts/datadog/api_update.go
```diff
@@ -0,0 +1,85 @@
...
+	// WorkspaceID is the workspace identifier (required).
+	WorkspaceID *string `json:"-"`
+}
+
+// Update updates the specified workspace alert.
+func Update(c *fastly.Client, i *UpdateInput) (*Alert, error) {
+	if i.AlertID == nil {
+		return nil, fastly.ErrMissingAlertID
+	}
+
```
anthony-gomez-fastly, line 39: missing checks for config being required

fastly/ngwaf/v1/workspaces/alerts/datadog/api_create.go
```diff
@@ -0,0 +1,73 @@
...
+	// Events is a list of event types (required).
+	Events []string `json:"events"`
+	// Type is the type of the workspace integration (required).
+	Type *string `json:"type"`
+	// WorkspaceID is the workspace identifier (required).
+	WorkspaceID *string `json:"-"`
+}
+
+// Create creates a new workspace alert.
```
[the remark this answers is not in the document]
rcaril: Do you mean we should have the integration name? 

ie "// Create creates a new datadog alert." ?

fastly/ngwaf/v1/workspaces/alerts/datadog/api_update.go
```diff
@@ -0,0 +1,85 @@
...
+	if i.WorkspaceID == nil {
+		return nil, fastly.ErrMissingWorkspaceID
+	}
+
+	// Get the current alert to validate the integration type.
+	currentAlert, err := Get(c, &GetInput{
```
anthony-gomez-fastly, line 45: what happens if we just try to update it without this check? @kpfleming should we do this check or let users potentially shoot themselves in the foot?

rcaril: The API calls are all the same, there aren't any type filters, which is why we had put this in place.

fastly/ngwaf/v1/workspaces/alerts/jira/api_create.go
```diff
@@ -0,0 +1,85 @@
...
+	// Config is the configuration associated with the workspace integration (required).
+	Config CreateConfig `json:"config"`
+	// Context, if supplied, will be used as the Request's context.
+	Context *context.Context `json:"-"`
+	// Description is an optional description for the alert (optional).
+	Description *string `json:"description"`
+	// Events is a list of event types (required).
+	Events []string `json:"events"`
+	// Type is the type of the workspace integration (required).
+	Type *string `json:"type"`
```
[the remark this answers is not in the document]
anthony-gomez-fastly: for create/update that is

fastly/ngwaf/v1/workspaces/alerts/datadog/api_update.go
kpfleming: Yeah, I think we should expose an API where each alert type is unique and they can't be converted from one type to another (even if the underlying API supports that).

fastly/ngwaf/v1/workspaces/alerts/jira/api_create.go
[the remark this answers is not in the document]
kpfleming: The type should not be exposed to the user of go-fastly at all, it should be hardcoded based on the package they have chosen.

fastly/ngwaf/v1/workspaces/alerts/datadog/api_test.go
```diff
@@ -0,0 +1,262 @@
...
+	"testing"
+
+	"github.com/fastly/go-fastly/v10/fastly"
+)
+
+// Global workspace value for the tests.
+var testWorkspaceID = fastly.TestNGWAFWorkspaceID
+
+func Test_Alerts(t *testing.T) {
+	t.Parallel()
```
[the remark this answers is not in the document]
rcaril: I didn't have any issues with the fixtures, are we saying this because there is a Create function involved?

fastly/ngwaf/v1/workspaces/alerts/datadog/api_create.go
```diff
@@ -0,0 +1,87 @@
...
+	// Description is an optional description for the alert.
+	Description *string
+	// Events is a list of event types (required).
+	Events *[]string
+	// WorkspaceID is the workspace identifier (required).
+	WorkspaceID *string
+}
+
+// Private struct to ensure correct alert type.
+type privateDatadogInput struct {
```
anthony-gomez-fastly, line 35: For each alert, since the user is defining what type of alert by specifying the package, there is a private input that has the `Type` field set to the correct type. all other fields are the same

fastly/ngwaf/v1/workspaces/alerts/datadog/api_update.go
```diff
@@ -0,0 +1,82 @@
...
+func Update(ctx context.Context, c *fastly.Client, i *UpdateInput) (*Alert, error) {
+	if i.AlertID == nil {
+		return nil, fastly.ErrMissingAlertID
+	}
+	if i.WorkspaceID == nil {
+		return nil, fastly.ErrMissingWorkspaceID
+	}
+
+	// Get the current alert to validate the integration type.
+	currentAlert, err := Get(ctx, c, &GetInput{
```
anthony-gomez-fastly, line 44: for update, we check that the alert is of the correct type before attempting to update

fastly/ngwaf/v1/workspaces/alerts/datadog/api_list.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	defer resp.Body.Close()
+
+	var was *Alerts
+	if err := json.NewDecoder(resp.Body).Decode(&was); err != nil {
+		return nil, fmt.Errorf("failed to decode json response: %w", err)
+	}
+
+	// Parse the alerts to only include the desired type of integration.
+	var parsedAlerts []Alert
```
anthony-gomez-fastly, line 48: we filter the alerts based on type for each list operation

philippschulte (changes requested):
This PR introduces the following new patterns:
1. Public and private input structs for `Create()`
2. Two calls for `Update()`

I would recommend the following:
1. Create an `AlertType` of type `string`
2. Set a `const` for each type of type `AlertType`
3. Add a type validator:

```
type AlertType string

const DataDogAlertType AlertType = "datadog"

func ValidateAlertType(s AlertType) bool {
    return s == DataDogAlertType
}
```
4. Add a type field to all input structs and validate in each function the the input has the correct type

That should avoid having more than one input structs for `Create()` and only one call for `Update()`.

However maybe I am missing something and the above is not an improvement. Let's think about this again or even ask Cameron Walters to weigh in on this.

fastly/ngwaf/v1/workspaces/alerts/datadog/api_create.go
anthony-gomez-fastly: the only thing we have that's similar is the thing i mentioned in your PR [here](https://github.com/fastly/go-fastly/pull/721#discussion_r2208577284). the reason we split them up is that both the config object for create/update and the response object are different for each type of alert

fastly/ngwaf/v1/workspaces/alerts/datadog/api_update.go
philippschulte: Please copy in the slack link where this solution was discussed for future reference. If Kevin has approved this approach then it is fine.

philippschulte (approved):

fastly/ngwaf/v1/workspaces/alerts/datadog/api_create.go
anthony-gomez-fastly: [Slack convo for reference](https://fastly.slack.com/archives/C08ESC78PGQ/p1751983232400109 )

PR #721, opened 2025-07-15
Combine workspace and account rules

philippschulte:
All Submissions:

### New Feature Submissions:

Does not pass test yet!

This is a breaking change!

### Notes for reviewers

The goal is to use one implementation for both account and workspace level rules. The same pattern can be applied for lists and signals. The necessary tooling is provided in this PR!

This PR should make it very do adopt the same pattern for lists and signals:
* remove all `WorkspaceID` fields from the input structs
* add `common.scope` to all input structs (the workspaceID needs to be added to the `applies_to` field inside the scope struct
* use `common.pathbuilder()` to set the proper path in each request function
* wrap the current test suite inside another a function which then gets call for a scope of "workspace" and "account"

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 
 ### Enhancements:
+- feat(ngwaf): add support for alerts ([#714](https://github.com/fastly/go-fastly/pull/714))
 
 - feat(ngwaf/v1/workspaces/thresholds): adds CRUD support for NGWAF Thresholds ([#713](https://github.com/fastly/go-fastly/pull/713))
 - feat(tls_custom_certificate): Add support for allow_untrusted_root attribute ([#596](https://github.com/fastly/go-fastly/pull/596))
--- a/fastly/alerts.go
+++ b/fastly/alerts.go
@@ -55,15 +55,15 @@ type AlertsMeta struct {
 
 // ListAlertDefinitionsInput is used as input to the ListAlertDefinitions function.
 type ListAlertDefinitionsInput struct {
-	// Cursor is the pagination cursor from a previous request's meta (optional).
+	// Cursor is the pagination cursor from a previous request's meta.
 	Cursor *string
-	// Limit is the maximum number of items included in each response (optional).
+	// Limit is the maximum number of items included in each response.
 	Limit *int
-	// Name filters definitions by name substring (optional).
+	// Name filters definitions by name substring.
 	Name *string
-	// ServiceID filters definitions by service (optional).
+	// ServiceID filters definitions by service.
 	ServiceID *string
-	// Sort is the field on which to sort definitions (optional).
+	// Sort is the field on which to sort definitions.
 	Sort *string
 }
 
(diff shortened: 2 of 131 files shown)


fastly/ngwaf/v1/common/pathbuilder.go
```diff
@@ -0,0 +1,40 @@
...
+// For scope type "account", the workspace ID is ignored.
+func BuildPath(scope *Scope, resource, resourceID string) (string, error) {
+	if scope == nil {
+		return "", fmt.Errorf("scope is required to build path")
+	}
+
+	switch scope.Type {
+	case ScopeTypeWorkspace:
+		if len(scope.AppliesTo) == 0 || scope.AppliesTo[0] == "" {
+			return "", fmt.Errorf("scope.applies_to must contain exactly one workspace ID for scope type 'workspace'")
```
anthony-gomez-fastly, line 22: this error doesn't explicitly check for `len(scope.AppliesTo) != 1`, should it?

fastly/ngwaf/v1/common/pathbuilder.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+
+	"github.com/fastly/go-fastly/v10/fastly"
+)
+
+// BuildPath generates the appropriate API path based on the given scope,
+// resource type, and resource ID.
+//
+// For scope type "workspace", the first element of AppliesTo is used as the workspace ID.
+// For scope type "account", the workspace ID is ignored.
```
anthony-gomez-fastly, line 13: > For scope type "account", `AppliesTo` is ignored.

philippschulte: Fixed in 0246f12784058497fc58a42ef0179ea28f2df030.

fastly/ngwaf/v1/rules/api_create.go
```diff
@@ -117,26 +107,31 @@ func Create(ctx context.Context, c *fastly.Client, i *CreateInput) (*Rule, error
 	}
 
 	v := struct {
-		Type           *string         `json:"type"`
+		Actions        []*CreateAction `json:"actions"`
```
anthony-gomez-fastly, line 110: When i did a similar thing in my PR, i made this an explicitly named private struct, do you have any strong feelings about that vs this approach?

anthony-gomez-fastly: I think i might change it to this approach, the named struct doesn't really accomplish anything

fastly/ngwaf/v1/rules/api_list.go
```diff
@@ -20,18 +21,17 @@ type ListInput struct {
 	Limit *int
 	// Page number of the collection to request.
 	Page *int
-	// Types filter results based on types (accepts more than one
-	// value and performs a union across rules of given types).
+	// Scope defines where the rule is applied, including its type (e.g., "workspace" or "account") and the specific IDs it applies to (required).
+	Scope *common.Scope
+	// Types filter results based on types (accepts more than one value and performs a union across rules of given types).
```
anthony-gomez-fastly, line 26: i think kevin changed all these comments to be shorter on purpose

philippschulte: Fixed in 5e826faf787f0f5cb98d7efe5defb7ef3286a927.

PR #725, opened 2025-07-29
fix(ngwaf/v1/workspaces): change some optional fields to omit empty

anthony-gomez-fastly:
All Submissions:

### New Feature Submissions:

### DIFF
--- a/fastly/ngwaf/v1/workspaces/api_response.go
+++ b/fastly/ngwaf/v1/workspaces/api_response.go
@@ -10,7 +10,7 @@ type Workspace struct {
 	AttackSignalThresholds *AttackSignalThresholds `json:"attack_signal_thresholds,omitempty"`
 	// ClientIPHeaders lists the request headers containing the
 	// client IP address.
-	ClientIPHeaders *[]string `json:"client_ip_headers,omitempty"`
+	ClientIPHeaders []string `json:"client_ip_headers,omitempty"`
 	// CreatedAt is the date and time in ISO 8601 format.
 	CreatedAt time.Time `json:"created_at"`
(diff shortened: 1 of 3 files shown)


fastly/ngwaf/v1/workspaces/api_response.go
```diff
@@ -7,10 +7,10 @@ import "time"
 type Workspace struct {
 	// AttackSignalThresholds are the parameters for system site
 	// alerts.
-	AttackSignalThresholds AttackSignalThresholds `json:"attack_signal_thresholds"`
+	AttackSignalThresholds *AttackSignalThresholds `json:"attack_signal_thresholds,omitempty"`
 	// ClientIPHeaders lists the request headers containing the
 	// client IP address.
-	ClientIPHeaders []string `json:"client_ip_headers"`
+	ClientIPHeaders *[]string `json:"client_ip_headers,omitempty"`
```
kpfleming, line 13: Slices are already pointers, so there's no need to make this a pointer. If there are no headers in the response, the slice will be empty.

philippschulte (commented):
Thanks for the update! Could you please provide some context on why these changes are necessary? It's currently unclear to me what problem they're solving. Some concrete examples or use cases would be helpful.

To my knowledge, we typically don’t omit fields from response structures. If that’s not the case, could you point to other resources in the codebase that follow a similar pattern for comparison?

I also noticed that you’re changing the types of some fields, which could introduce a breaking change. If there's a way to avoid that—especially right now—that would be preferable.

Lastly, keep in mind that Go 1.24 introduces the omitzero struct tag, which might be relevant for what you're trying to achieve.

Thanks!

PR #755, opened 2025-09-12
breaking(ngwaf/rules): add support for the multival condition type

rcaril:
### Change summary

This PR introduces support for the `multival` condition type for NGWAF rules. 

### New Feature Submissions:

### Changes to Core Features:

New structs were created for multival to ensure Go comments were accurate. 

Tests:
```
ok      github.com/fastly/go-fastly/v11/fastly/ngwaf/v1/rules   17.176s
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 [Full Changelog](https://github.com/fastly/go-fastly/compare/v11.2.0...)
 
 ### Breaking:
+- breaking(ngwaf/rules): group and multival conditions no longer accept a 'type' field
 
 ### Enhancements:
 - feat(ngwaf/rules): add support for the multival condition type ([#755](https://github.com/fastly/go-fastly/pull/755))


fastly/ngwaf/v1/rules/api_create.go
```diff
@@ -88,6 +91,22 @@ type CreateCondition struct {
 	Value *string `json:"value"`
 }
 
+// CreateConditionMult defines a multival condition.
+type CreateConditionMult struct {
+	// Type specifies the condition type (must be "single")
+	// (required).
+	Type *string `json:"type"`
```
anthony-gomez-fastly, line 98: so this can only be `single`? if so we should just hide it and always set it to single

rcaril: Addressed in [3f543d7](https://github.com/fastly/go-fastly/pull/755/commits/3f543d7f04ab07704086c4cd7693d7b7ff8dfd9b).

fastly/ngwaf/v1/rules/api_create.go
```diff
@@ -126,6 +159,32 @@ type CreateClientIdentifier struct {
 	Type *string `json:"type"`
 }
 
+// MarshalJSON implements json.Marshaler to automatically include the required
+// "type": "single" field during JSON serialization.
+func (c *CreateConditionMult) MarshalJSON() ([]byte, error) {
```
anthony-gomez-fastly, line 164: i think we use either internal or anonymous structs in the other places we do this, check out `fastly/ngwaf/v1/rules/api_create.go`

rcaril: As per our chat, we are going to implement the same pattern used in `fastly/ngwaf/v1/workspaces/alerts/datadog/api_create.go`.

kpfleming (approved):

PR #756, opened 2025-09-15
Add new workflow to check for golangci-lint releases

philippschulte:
### Change summary

It's important to keep up with the latest and greatest of [golangci-lint](https://github.com/golangci/golangci-lint) in order to prevent breaking changes for linting issues as experienced in #754.

This PR adds a new workflow/cron job which checks for new releases every hour. If a new release is detected the workflow works as follows:

- creates a new GitHub issue titled 🚀 New golangci-lint release: vX.Y.Z
- provides a link and release notes in the body

That issue stays open until a team member opens a pull request with a  dependency bump which references it and closes it automatically.

If we as a team agree on this new workflow the plans is to transfer the same workflow to all of our `Go` projects which use [golangci-lint](https://github.com/golangci/golangci-lint).

 All Submissions:

### DIFF
--- a/.github/workflows/check-golangci-release.yml
+++ b/.github/workflows/check-golangci-release.yml
@@ -1,0 +1,52 @@
+name: Notify on golangci-lint Release
+
+on:
+  schedule:
+    - cron: "0 * * * *" # every hour
+  workflow_dispatch:
+jobs:
+  check-release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Get latest golangci-lint release
+        id: get_latest
+        run: |
+          release=$(curl -s https://api.github.com/repos/golangci/golangci-lint/releases/latest)
+          tag=$(echo "$release" | jq -r .tag_name)
+          url=$(echo "$release" | jq -r .html_url)
+          body=$(echo "$release" | jq -r .body)
+          echo "tag=$tag" >> $GITHUB_OUTPUT
+          echo "url=$url" >> $GITHUB_OUTPUT
+          echo "body<<EOF" >> $GITHUB_OUTPUT
+          echo "$body" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+      - name: Compare with last notified release
+        id: compare
+        run: |
+          last=$(cat .last_release 2>/dev/null || echo "")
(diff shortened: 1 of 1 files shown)


philippschulte:
> i think since we are using the golangci-lint github action dependabot will automatically check for new versions of that? 

Since we’re using the `golangci-lint` GitHub Action, you might expect Dependabot to automatically update us when new versions are released. In practice, that hasn’t been happening—otherwise we would already be up to date.

> i would also try to only run this once a day, once an hour seems a bit much if we do go forward with it

Regarding the schedule: running this workflow once per hour might seem frequent, but it doesn’t cost us anything. The benefit of checking hourly is that we avoid situations like #754, where we fell behind on updates. If a new release issue is created, we can treat it as a priority before merging other PRs that day.

Also worth noting: the workflow will only ever create **one issue per release**, so the frequency of the cron job won’t result in duplicates.

I’m in favor of having a dedicated workflow for this and running it hourly, but I’m open to other suggestions if the team feels differently.

philippschulte:
@anthony-gomez-fastly  yes, that's a good idea with the `makefile`! I haven't thought about that! I think this is a good plan to move forward instead of adding a new pr and manually bumping all repos every time a new release is out. And not to forget to keep the local instance in sync.

kpfleming:
Dependabot can't handle this for us because the golangci-lint Action has a separate versioning scheme from golangci-lint itself; the Action is currently at version 8.0.0 while the tool is at version 2.4.0. Dependabot will tell us about new versions of the Action, but not new versions of the tool.

The only way that Dependabot could tell us about new versions of the tool would be for it to be listed in `go.mod` or a similar file, but we've already determined that we don't want to do that.

I think the best course of action may be to pin the version of golangci-lint in every place that we use it (including in `Makefile` so that local builds use the same version), but add an additional job in the Actions workflow which allows the Action to use the `latest` version. That job's status would not be required for a PR to be mergeable, it would just be informational, but it would mean that every PR would be tested against both the pinned and `latest` versions and if a new release of golangci-lint starts causing failures we'll see them the next time a PR's Actions workflows are run. We can then decide when/how we want to address the failures in a separate PR which also changes the pinned version to the new version.

philippschulte:
I'll close this one out in favor of #757.

PR #764, opened 2025-10-28
Support `max_use` and `max_lifetime` backend parameters

ulyssa:
### Change summary

This is a draft PR for adding support for two different backend fields that are going to be added to the API:

- `max_use`, which specifies how many times an HTTP keepalive connection can be used before it's no longer pooled.
- `max_lifetime`, which specifies in milliseconds an upper bound for how long an HTTP keepalive connection can be open before it's no longer used.

 All Submissions:

### New Feature Submissions:

    * Yes

### Changes to Core Features:

* [ ] Have you written new tests for your core changes, as applicable?
    * I didn't add any tests, since this just updates what can be passed in the struct to the API.
    * Yes

### User Impact

This will allow using a newly exposed feature in both Delivery and Compute, which have added support for honoring these backend parameters in their connection pooling logic.

### DIFF
--- a/fastly/backend.go
+++ b/fastly/backend.go
@@ -22,6 +22,8 @@ type Backend struct {
 	Hostname            *string    `mapstructure:"hostname"`
 	KeepAliveTime       *int       `mapstructure:"keepalive_time"`
 	MaxConn             *int       `mapstructure:"max_conn"`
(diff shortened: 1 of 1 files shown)


rcaril:
Being addressed in https://github.com/fastly/go-fastly/pull/807.

PR #805, opened 2026-05-01
[CDTOOL-1321] Add DNS Zones and TSIG Keys Support

rcaril:
### Change summary

This PR adds support for Fastly DNS Zones and TSIG Keys. 

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### Notes:

 Added a generic `Nullable[T]` helper to `fastly/helpers.go` to support fields that need to distinguish between "omit from request" and "explicitly set to null". This was specifically needed to support the `inbound_tsig_key_id` field. 
 Internal conversation on this can be found [here](https://fastly.slack.com/archives/C09MAG96C2D/p1778004212920419).


fastly/errors.go
```diff
@@ -83,6 +83,10 @@ var ErrMissingACLID = NewFieldError("ACLID")
 // requires a "Address" key, but one was not set.
 var ErrMissingAddress = NewFieldError("Address")
 
+// ErrMissingAlgorithm is an error that is returned when an input struct
+// requires a "Algorithm" key, but one was not set.
+var ErrMissingAlgorithm = NewFieldError("Algorithm")
+
```
jedisct1: Minor but since other errors are lexicographically sorted, maybe that one should be after `ErrMissingAlertID`?

fastly/dns/v1/tsigkeys/api_update.go
```diff
@@ -0,0 +1,45 @@
...
+	Algorithm *string `json:"algorithm,omitempty"`
+	// Secret is the Base64 encoded secret key.
+	Secret *string `json:"secret,omitempty"`
+}
+
+// Update updates an existing TSIG key.
+func Update(ctx context.Context, c *fastly.Client, i *UpdateInput) (*TSIGKey, error) {
+	if i.TSIGKeyID == nil {
+		return nil, fastly.ErrMissingID
+	}
```
jedisct1, line 29: Is it possible to change only the algorithm without providing a corresponding secret? Maybe the API accepts this, but that wouldn't really make sense.

fastly/dns/v1/dnszones/api_response.go
```diff
@@ -0,0 +1,75 @@
...
+
+// Zone is the API response structure for the create, update and get operations.
+type Zone struct {
+	// ID is the zone identifier (UUID).
+	ID *string `json:"id"`
+	// Name is the domain name for your zone.
+	Name *string `json:"name"`
+	// Description is a freeform descriptive note.
+	Description *string `json:"description"`
+	// Type is the type of the zone.
```
jedisct1, line 13: "Type is the type" sounds a little bit recursive :)

Maybe list the possible values (`"primary"` and `"secondary"` ?)

fastly/dns/v1/dnszones/api_response.go
```diff
@@ -0,0 +1,75 @@
...
+	// NotifyIPAddresses are IP addresses where Primary DNS servers can send NOTIFY messages.
+	NotifyIPAddresses *NotifyIPAddresses `json:"notify_ip_addresses"`
+	// InboundTSIGKeyID is the ID of the TSIG key used to secure inbound zone transfers.
+	InboundTSIGKeyID *string `json:"inbound_tsig_key_id"`
+}
+
+// XfrConfigInboundInput specifies the inbound zone transfer attributes for create and update requests.
+type XfrConfigInboundInput struct {
+	// Primaries is an array of primary DNS server objects for inbound zone transfers.
+	Primaries []Primary `json:"primaries,omitempty"`
```
jedisct1, line 40: There's no way to distinguish "don't change primaries" from "clear primaries" : both `nil` and `[]Primary{}` are represented the same way.

Is it intentional that callers can't clear the primaries via `Update()`?

fastly/dns/v1/dnszones/api_response.go
```diff
@@ -0,0 +1,75 @@
...
+	// InboundTSIGKeyID is the ID of the TSIG key used to secure inbound zone transfers.
+	InboundTSIGKeyID *string `json:"inbound_tsig_key_id,omitempty"`
```
jedisct1: There's no way to send an explicit `null` (`nil` would just omit the field).

How to clear the TSIG Key ID? (presumably optional, looking at the fixtures) Does the API clear it when it receives an empty string, or does it require `null`?

jedisct1 (changes requested):

fastly/dns/v1/dnszones/api_response.go
rcaril: I agree it may sound recursive, but this is the terminology from the API docs. We have this type of pattern across the codebase. 
IE:
https://github.com/fastly/go-fastly/blob/5df365ec97f2af08a284bfaf53cd9c15ca40aabc/fastly/notifications.go#L90 
https://github.com/fastly/go-fastly/blob/5df365ec97f2af08a284bfaf53cd9c15ca40aabc/fastly/service_details.go#L116

> Maybe list the possible values ("primary" and "secondary" ?)

The only possible value is `secondary`. However, we normally provide the possible values downstream rather then in go-fastly.

fastly/dns/v1/tsigkeys/api_update.go
rcaril: I tested this, it is possible. However we don't know what the API is doing under the hood - perhaps it's re-hashing the secret. go-fastly is only a wrapper, so we shouldn't enforce anything outside of the API spec.

fastly/errors.go
rcaril: Fixed in [b0b4284](https://github.com/fastly/go-fastly/pull/805/commits/b0b4284aaadd9fa0619981346104df6daabe4c39)[b0b4284](https://github.com/fastly/go-fastly/pull/805/commits/b0b4284aaadd9fa0619981346104df6daabe4c39).

jedisct1: Thank you!

fastly/dns/v1/dnszones/api_response.go
rcaril: I believe so, I don't see anything in the API spec that specifies that the field should be `nullable`. Setting an existing attribute to `nil` isn't an existing pattern in `go-fastly`.

fastly/dns/v1/dnszones/api_response.go
rcaril: Started an internal chat on this here:
https://fastly.slack.com/archives/C09MAG96C2D/p1778004212920419

rcaril: Added support for setting the field to `null` in [e0f436b](https://github.com/fastly/go-fastly/pull/805/commits/e0f436bdddc4dec9a91c4e6a2992f66d2eade5d0). The expected payload is:
```
{"xfr_config_inbound":{"inbound_tsig_key_id":null}}'
```

fastly/dns/v1/dnszones/api_response.go
rcaril: Confirmed this is expected [here](https://fastly.slack.com/archives/C09MAG96C2D/p1778012304936699?thread_ts=1778004212.920419&cid=C09MAG96C2D).

rcaril:
Since I added a new helper function i'm looping in @kpfleming to eyeball this PR.

fastly/dns/v1/tsigkeys/api_update.go
kpfleming: No, in DNS itself this is not possible, as the secret is generated using the specified algorithm.

rcaril: > No, in DNS itself this is not possible, as the secret is generated using the specified algorithm.

Should we require both fields to be updated in the scenario that the algorithm field is modified?

fastly/dns/v1/dnszones/api_list.go
```diff
@@ -0,0 +1,56 @@
...
+
+	"github.com/fastly/go-fastly/v14/fastly"
+)
+
+// ListInput specifies the information needed to list zones.
+type ListInput struct {
+	// Cursor is the value from the next_cursor field of a previous
+	// response, used to retrieve the next page. To request the first
+	// page, this should be empty.
+	Cursor *string
```
kpfleming: I could be mistaken, but I remembered the pattern being that go-fastly handles the pagination and it isn't exposed to the user of go-fastly at all: they just get the list back. We have various helpers to handle different types of pagination, hopefully we have one for cursor-based pagination.

fastly/dns/v1/tsigkeys/api_update.go
kpfleming: No, leave that up to the API to handle.

fastly/dns/v1/dnszones/api_list.go
rcaril: Good call out, let me compare this to our recent Logging Error feature.

rcaril: Corrected pagination in [a2fb351](https://github.com/fastly/go-fastly/pull/805/commits/a2fb35120a0ed2e7aa557c0fddfe40cc0fbaf560).

jedisct1 (approved):

kpfleming (approved):

PR #813, opened 2026-05-22
feat(contact): add support for customer contacts

bclerc:
### Change summary

Adds support for the [customer contacts API](https://www.fastly.com/documentation/reference/api/account/contact/) under `/customer/{customer_id}/contacts`.

Three new client methods:

- `Client.ListContacts`
- `Client.CreateContact`
- `Client.DeleteContact`

The endpoint uses JSON:API with the `customer_contact` resource type. The Fastly API does not expose an update endpoint for contacts, so only list/create/delete are implemented.

The integration test creates a guard contact first because the API refuses to delete the last contact of a given type (`You cannot remove the last <type> contact`).

All Submissions:

- [x] Have you followed the guidelines in our Contributing document?
- [x] Have you checked to ensure there aren't other open Pull Requests for the same update/change?

### New Feature Submissions:

- [x] Does your submission pass tests?

### User Impact

Users can now manage customer-level contacts (primary, billing, technical, security, emergency) through the Go client instead of having to call the API directly. This unblocks support for customer contacts in the Terraform provider.

No breaking changes. New surface only.


kpfleming:
Thanks for the PR! Unfortunately we are no longer adding API endpoints to the top-level `fastly` package unless they belong there because there are existing endpoints for the same resource type. Since this is a new resource type, it will need to be put into `fastly/customer/contacts` - refer to `fastly/apisecurity/operations` for an example of how this is done.

bclerc:
Thanks for the quick review and for pointing me to the right pattern! I'll move the new endpoints into a dedicated `fastly/customer/contacts` sub-package and use `fastly/apisecurity/operations` as the reference. I'll push the refactor shortly.

bclerc:
Done — I've pushed the refactor. The contacts client now lives at `fastly/customer/contacts` following the `fastly/apisecurity/operations` layout (`Create`/`List`/`Delete` with `(ctx, *fastly.Client, *Input)` signatures, pointer-valued inputs, fixtures moved under the sub-package). The endpoint is JSON:API so the code imports `github.com/google/jsonapi` directly for `MediaType` and the unmarshaling helpers — happy to adjust if you'd prefer a different approach there. CHANGELOG entry updated to `feat(customer/contacts)`.

kpfleming:
Thanks - we're about to start a three day weekend, so we'll look into this next week.

fastly/customer/contacts/api_create.go
```diff
@@ -0,0 +1,88 @@
...
+	}
+	return &co, nil
+}
+
+func deref(s *string) string {
+	if s == nil {
+		return ""
+	}
+	return *s
+}
```
kpfleming: I believe `ToValue()` already exists for this purpose.

fastly/customer/contacts/api_contacts_test.go
```diff
@@ -0,0 +1,93 @@
...
+	)
+	fastly.Record(t, "get_current_user", func(c *fastly.Client) {
+		var u *fastly.User
+		u, err = c.GetCurrentUser(ctx)
+		if err == nil && u != nil && u.CustomerID != nil {
+			customerID = *u.CustomerID
+		}
+	})
+	require.NoError(t, err)
+	require.NotEmpty(t, customerID)
```
kpfleming: This does not need to be recorded as a fixture, it's an API call which is not part of the package being tested.

kpfleming (changes requested):
Some small change requests, but overall this look great!

fastly/customer/contacts/api_create.go
bclerc: Thanks, I'd missed that one. Updated to use `fastly.ToValue` and dropped the local helper.

fastly/customer/contacts/api_contacts_test.go
bclerc: Good point. Removed the `GetCurrentUser` call and its fixture. I now source the customer ID from a hardcoded constant in the test, the same way `apisecurity/operations` uses `fastly.TestDeliveryServiceID` for its service ID.

fastly/customer/contacts/api_contacts_test.go
```diff
@@ -12,19 +12,11 @@ import (
...
-			customerID = *u.CustomerID
-		}
-	})
-	require.NoError(t, err)
-	require.NotEmpty(t, customerID)
+	var err error
+
+	// The customer ID is an opaque identifier that is only replayed against the
+	// recorded fixtures; its exact value does not matter for the test.
+	customerID := "7i6ZbMEdjnTUNyk75XgWO0"
```
kpfleming, line 19: This should probably use the `GetCurrentCustomer` API endpoint (https://www.fastly.com/documentation/reference/api/account/customer/#get-logged-in-customer), but without recording it into the fixtures. This can be done by using `DefaultClient()` to get a regular Fastly client, avoiding the recording process.

kpfleming (changes requested):

fastly/customer/contacts/api_contacts_test.go
bclerc: Sorry for the delay, busy week. GetCurrentCustomer isn't implemented in go-fastly today, so i'm not sure i follow what you mean here. I hardcoded the customer id because that's how it's done in the other tests (token_test.go, account_event_test.go), so i followed the same pattern.
If you'd rather fetch it at runtime, the closest thing i see is to grab a regular client with DefaultClient() and call GetCurrentUser, the returned User have a CustomerID field, something like:

```go  
user, err := fastly.DefaultClient().GetCurrentUser(ctx)
require.NoError(t, err)
customerID := fastly.ToValue(user.CustomerID)
 ```

 Let me know which you prefer.

kpfleming: The underlying issue is that hardcoding a CID won't work, because it will have to match the CID linked to the token that is being used to authenticate to the API or the tests will fail. If we followed the existing pattern there would have to be additions to `fastly_test_utils.go` to hold the hardcoded CID, and every developer trying to run this test would have to provide the CID which matches the token they are using.

However, since these are the only API calls/tests which require a CID, I would be fine with determining that CID at runtime so that running the tests doesn't require additional setup. While `GetCurrentUser` can be used for that, we've also seen a number of cases where that API operation returns an error because the API token is service-limited or has some other limits on it, but `GetCurrentCustomer` should not fail in those cases. That's why I suggested using `GetCurrentCustomer`.

We can implement `GetCurrentCustomer` if that's something you'd prefer us to do. it would be implemented in the `fastly` package since it's a client-related operation and does not need to be in a subpackage.

bclerc: Heads up: I'll have limited availability over the next couple of weeks. I'll get back to this PR as soon as I'm able. Thanks for your patience!

PR #815, opened 2026-05-28
Add 'zizmor' static analyzer for Actions workflows

kpfleming:
All findings have been addressed except for 'insufficient cooldown' in the Dependabot configuration, since that needs a team-level discussion.


jedisct1 (approved):
LGTM!

PR #827, opened 2026-06-15
[CDTOOL-1086] Add support for listing Time Series metrics

rcaril:
### Change summary

This PR adds support for the `GET` `/ngwaf/v1/timeseries` [endpoint](https://www.fastly.com/documentation/reference/api/ngwaf/timeseries/#ngwafListTimeseries). 

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

Help docs for the existing workspace `GET` Time Series endpoint have also been updated to reflect the latest API docs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 ### Enhancements:
 
-- feat(ngwaf/timeseries): add support for list time series metrics ([#827](https://github.com/fastly/go-fastly/pull/827))
+- feat(ngwaf/timeseries): add support for listing time series metrics ([#827](https://github.com/fastly/go-fastly/pull/827))
 
 ### Dependencies:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Enhancements:
 
+- feat(ngwaf/timeseries): add support for list time series metrics ([#827](https://github.com/fastly/go-fastly/pull/827))
+
 ### Dependencies:
 
 ### Bug fixes:
--- a/fastly/ngwaf/v1/timeseries/api_list.go
+++ b/fastly/ngwaf/v1/timeseries/api_list.go
@@ -9,48 +9,45 @@ import (
 	"github.com/fastly/go-fastly/v15/fastly"
 )
 
-// GetInput specifies the information needed for the Get() function to
+// ListInput specifies the information needed for the List() function to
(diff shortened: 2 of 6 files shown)


fastly/ngwaf/v1/timeseries/api_list.go
```diff
@@ -0,0 +1,68 @@
...
+	// timeseries. Metrics can be XSS, SQLI, HTTP404, requests_total,
+	// requests_attack, requests_total_blocked, or any custom metric (required).
+	Metrics *string
+	// To is the end of a date-time range, expressed in RFC 3339 format.
+	To *string
+}
+
+// List retrieves timeseries metrics for Next-Gen WAF.
+func List(ctx context.Context, c *fastly.Client, i *ListInput) (*Timeseries, error) {
+	if i.Metrics == nil {
```
kpfleming: This isn't sufficient, as the struct field could be non-nil but point to an empty string, and the API call will return a failure response. If we're going to validate the input for 'required' fields, we'll need to check that too.

The simplest way is to just change these to plain `string` instead of `*string`, and then check if they are empty. Since they are required there isn't any benefit in allowing the caller to provide `nil`, a pointer to an empty string, or a pointer to a non-empty string: the first two are both validation failures.

However... the other APIs in this family don't work that way, so the solution here will have to be to check for nil and then check for non-empty.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Enhancements:
 
+- feat(ngwaf/timeseries): add support for list time series metrics ([#827](https://github.com/fastly/go-fastly/pull/827))
```
kpfleming:
```suggestion
- feat(ngwaf/timeseries): add support for listing time series metrics ([#827](https://github.com/fastly/go-fastly/pull/827))
```

kpfleming (changes requested):
Some small changes, otherwise this looks good!

fastly/ngwaf/v1/timeseries/api_list.go
rcaril: Ah, thanks for catching this! Updated in [d60fdc7](https://github.com/fastly/go-fastly/pull/827/commits/d60fdc777b57848f1cc9cca48073275db683a195) to add the second check for the non-empty string.

kpfleming (approved):

PR #829, opened 2026-06-17
Add required fields for custom TLS

caiges:
### Change summary

Add creation and deletion of custom TLS configurations. This is building up changes required to fully support custom TLS configurations in the Fastly Terraform Provider.

 

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

There shouldn't be any impact to existing users.

- [ ] Some of this functionality may require Fastly to enable for account-specific features.
- [ ] Public type structure may not be desired shape.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
 ### Enhancements:
 
 - feat(ngwaf/timeseries): add support for listing time series metrics ([#827](https://github.com/fastly/go-fastly/pull/827))
+- feat(tls/custom-configuration): add `CreateCustomTLSConfiguration` and `DeleteCustomTLSConfiguration`, and surface `default_certificate`, `default_ecdsa_certificate`, `tls_1_2_cipher_suite_profile`, `tls_1_3_cipher_suite_profile`, and `vipspace` on the configuration resource. Create input also accepts `cipher_suites`.
+
+### Bug fixes:
 
 ### Dependencies:
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Enhancements:
 
 - feat(ngwaf/timeseries): add support for listing time series metrics ([#827](https://github.com/fastly/go-fastly/pull/827))
-- feat(tls/custom-configuration): add `CreateCustomTLSConfiguration` and `DeleteCustomTLSConfiguration`, and surface `default_certificate`, `default_ecdsa_certificate`, `tls_1_2_cipher_suite_profile`, `tls_1_3_cipher_suite_profile`, and `vipspace` on the configuration resource. Create input also accepts `cipher_suites`.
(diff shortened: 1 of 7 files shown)


anthony-gomez-fastly (approved):
LGTM! thank you!

anthony-gomez-fastly:
@caiges looks great, please add a changelog entry

caiges:
@anthony-gomez-fastly Added alongside the other unreleased change.

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
 ### Enhancements:
 
 - feat(ngwaf/timeseries): add support for listing time series metrics ([#827](https://github.com/fastly/go-fastly/pull/827))
+- feat(tls/custom-configuration): add `CreateCustomTLSConfiguration` and `DeleteCustomTLSConfiguration`, and surface `default_certificate`, `default_ecdsa_certificate`, `tls_1_2_cipher_suite_profile`, `tls_1_3_cipher_suite_profile`, and `vipspace` on the configuration resource. Create input also accepts `cipher_suites`.
```
anthony-gomez-fastly:
```suggestion
- feat(tls/custom-configuration): add `CreateCustomTLSConfiguration`, `DeleteCustomTLSConfiguration`, and additional supported fields. ([#829](https://github.com/fastly/go-fastly/pull/829))
```

kpfleming:
Unfortunately we cannot make these changes as they are, because these fields are only available to a subset of customers who have additional TLS platform features enabled on their accounts. The API endpoints are the same for both normal and enhanced TLS features, but the go-fastly functionality cannot be the same because the API will reject requests that include attributes which are not supported by the normal accounts. This is an artifact of an API design decision which was made long ago and predates everyone involved today :-)

For this to be supported, go-fastly (and the CLI/Terraform provider) will have to have a parallel set of functions which are only used by customers who have access to the enhanced TLS features. In addition since these features are not part of our public documentation (at least that was true the last time I checked), we technically should not even support them in go-fastly. That part of the discussion will need to involve the TLS engineering team and our Product Management team.

caiges:
> This is an artifact of an API design decision which was made long ago and predates everyone involved today :-)

:laughing: I am all too familiar with this phenomenon.

> we technically should not even support them in go-fastly. That part of the discussion will need to involve the TLS engineering team and our Product Management team.

My work here and in a fork of the provider was based on these fields being optional but semi-private fields tend to become the public API in practice so can also understand this too.

kpfleming:
If you want to pursue this that would certainly be welcome, but as I mentioned previously the approach would have to be creating a new set of API functions (which use the `SPP` label to distinguish them from the regular functions) so that the existing functions continue to operate as they do now.

kpfleming:
I'm going to close this but we won't forget the desire to do this. We've been having discussions internally and it appears that Real Soon Now the provider will be able to query 'feature flags' on customer accounts so it will be possible to ensure that the correct resources and API endpoints are used.

caiges:
@kpfleming thanks for the update. I'm more than happy to take another swing at this once there's more movement there.

PR #839, opened 2026-07-23
[Breaking] Allow Logging `placement` to be set to `null` for delivery services

rcaril:
### Change summary

This PR updates the `placement` attribute on all logging endpoints so it can be set to `null`, which restores automatic placement of the logging call in the generated VCL. Previously, if a user set `placement` to the value of  `none` (suppressing the logging call), it was not possible to reset it back to `null`.

See [API docs](https://www.fastly.com/documentation/reference/api/logging/datadog/) for full details on use case. 

 All Submissions:

### New Feature Submissions:

### Changes to Core Features:

### User Impact

Users will be able to change the `placement` logging attribute back to `null` if it was previously was set to `none`. 

`Placement` on the `Update*Input` structs changes from `*string` to `*fastly.Nullable[string]`, and updates now send JSON instead of form encoding.

Compute / WASM services only accept the value of `none` for `placement`, where VCL / Delivery services need the ability to reset to `null`.


fastly/logging_bigquery_test.go
```diff
@@ -216,7 +220,7 @@ func TestClient_Bigqueries(t *testing.T) {
 			AccountName:    ToPointer("service-account"),
 			SecretKey:      ToPointer(secretKey),
 			Format:         ToPointer("{\n \"timestamp\":\"%{begin:%Y-%m-%dT%H:%M:%S}t\",\n  \"time_elapsed\":%{time.elapsed.usec}V,\n  \"is_tls\":%{if(req.is_ssl, \"true\", \"false\")}V,\n  \"client_ip\":\"%{req.http.Fastly-Client-IP}V\",\n  \"geo_city\":\"%{client.geo.city}V\",\n  \"geo_country_code\":\"%{client.geo.country_code}V\",\n  \"request\":\"%{req.request}V\",\n  \"host\":\"%{req.http.Fastly-Orig-Host}V\",\n  \"url\":\"%{json.escape(req.url)}V\",\n  \"request_referer\":\"%{json.escape(req.http.Referer)}V\",\n  \"request_user_agent\":\"%{json.escape(req.http.User-Agent)}V\",\n  \"request_accept_language\":\"%{json.escape(req.http.Accept-Language)}V\",\n  \"request_accept_charset\":\"%{json.escape(req.http.Accept-Charset)}V\",\n  \"cache_status\":\"%{regsub(fastly_info.state, \"^(HIT-(SYNTH)|(HITPASS|HIT|MISS|PASS|ERROR|PIPE)).*\", \"\\\\2\\\\3\") }V\"\n}"),
-			Placement:      ToPointer("waf_debug"),
```
rcaril, line 219: `waf_debuf` is a legacy value that is no longer in use.

fastly/logging_kafka_test.go
```diff
@@ -216,11 +216,15 @@ func TestClient_Kafkas(t *testing.T) {
 			NewName:          ToPointer("new-test-kafka"),
 			Topic:            ToPointer("new-kafka-topic"),
 			ProcessingRegion: ToPointer("eu"),
+			Placement:        NullValue[string](),
```
philippschulte, line 219: Non-blocking: Since this update now uses JSON, could we also exercise `ParseLogKeyvals` and `UseTLS` here to verify that `Compatibool` still serializes correctly?

philippschulte (approved):
Looks good overall. The nullable placement behavior is implemented consistently across the logging endpoints, and the tests cover resetting delivery services to automatic placement while preserving `none` for Compute services.

Thank you!

fastly/logging_kafka_test.go
rcaril: Updated in [af1d781](https://github.com/fastly/go-fastly/pull/839/commits/af1d7816e554718d87eaaf8053bf2d92263bc8af) - `Kafka`, `Logentry`, `Splunk` and `Syslog` have been updated with these additioanl serialization checks.

philippschulte (approved):
Thank you!

PR #846, opened 2026-08-04
[CDTOOL-1692] add support for audit log event mappings

anthony-gomez-fastly:
### Change summary

[API docs](https://www.fastly.com/documentation/reference/api/observability/notifications/event-mappings/)

Add CRUD support for creating, listing, getting, updating, and deleting audit log event mappings, plus discovery endpoints for supported event types and scope types, via the Notification Service's Event Mappings API.

- Add fastly/notifications/v1/eventmappings package (Create, Get, List, Update, Delete)
- Add fastly/notifications/v1/eventmappings/eventtypes package (List)
- Add fastly/notifications/v1/eventmappings/scopetypes package (List)

 All Submissions:

### New Feature Submissions:


fastly/notifications/v1/eventmappings/api_list.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/fastly/go-fastly/v17/fastly"
+)
+
+// ListInput specifies the information needed for the List() function to
+// perform the operation.
+type ListInput struct {
+	// Cursor is the pagination cursor from the NextCursor field of a
+	// previous response, used to retrieve the next page. To request the
+	// first page, this should be nil.
+	Cursor *string
```
kpfleming: In many API 'list' operations in go-fastly we hide the pagination and return a slice of results, so the caller does not need to deal with pagination. We don't have a standard way to decide whether or not to do that, but I think most of the time we do hide the pagination from the caller.

kpfleming (dismissed):
Only one comment related pagination in the 'list' operation for event mappings. Otherwise this looks very good.

fastly/notifications/v1/eventmappings/api_list.go
anthony-gomez-fastly: fixed in `fb919182`

fastly/notifications/v1/eventmappings/api_list.go
```diff
@@ -40,19 +33,38 @@ type ListInput struct {
...
+		out    []EventMapping
+		cursor *string
+	)
+	for {
+		page, err := listPage(ctx, c, i, cursor)
+		if err != nil {
+			return nil, err
+		}
+		out = append(out, page.Data...)
+		if page.Meta.NextCursor == nil || *page.Meta.NextCursor == "" {
```
kpfleming, line 49: I would have bet we already had a cursor-based pagination helper in go-fastly, but I don't see one!

kpfleming (approved):

fastly/notifications/v1/eventmappings/api_list.go
anthony-gomez-fastly: i also thought that 🤷

PR #849, opened 2026-08-07
Upgrade Golang to 1.26.5

anthony-gomez-fastly:
### Change summary

Bump go.mod and CI (setup-go, golangci-lint) from 1.25 to 1.26.5. golangci-lint is pinned to v2.12.2 since the previous v2.10.1 pin predates 1.26 support.

Go 1.26's stricter go vet printf check now flags %q used with int-typed arguments (previously silently accepted), which broke `go test`.

 All Submissions:

### New Feature Submissions:


rcaril (approved):
LGTM!

PR #851, opened 2026-08-13
[CDTOOL-1698] Add Log Explorer and Insights API support

philippschulte:
### Change summary

Adds go-fastly client and VCR test support for the Log Explorer and Log Insights read-only endpoints.

 All Submissions:

* [x] Have you followed the guidelines in our Contributing document?
* [x] Have you checked to ensure there aren't other open [Pull Requests](https://github.com/fastly/go-fastly/pulls) for the same update/change?

### New Feature Submissions:

* [x] Does your submission pass tests?

PR #853, opened 2026-08-17
[CDTOOL-1699] Improve Log Explorer and Insights consumer support

philippschulte:
### Change summary

Follow-up to #851.

A live Log Insights request using `domain` and `pops` confirmed that both values are echoed in `meta.filters`, but the current `LogInsightsFilters` response model does not expose them. This change adds those response fields so consumers can access the complete metadata returned by the API.

It also exports the supported Log Explorer filter fields/operators and Log Insights visualizations. This allows consumers such as the CLI to build validated flag choices and a better CLI experience without duplicating the supported API values locally.

 All Submissions:

### New Feature Submissions:

### User Impact

N/A

### DIFF
--- a/fastly/observability_log_explorer.go
+++ b/fastly/observability_log_explorer.go
@@ -35,6 +35,21 @@ const (
 	LogExplorerFilterFieldFastlyIsCacheHit LogExplorerFilterField = "fastly_is_cache_hit"
 )
 
+// LogExplorerFilterFields is a list of supported Log Explorer filter fields.
+var LogExplorerFilterFields = []LogExplorerFilterField{
+	LogExplorerFilterFieldDomain,
+	LogExplorerFilterFieldRequestPath,
+	LogExplorerFilterFieldFastlyPOP,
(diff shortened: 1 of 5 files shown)


fastly/observability_log_explorer.go
```diff
@@ -35,6 +35,21 @@ const (
 	LogExplorerFilterFieldFastlyIsCacheHit LogExplorerFilterField = "fastly_is_cache_hit"
 )
 
+// LogExplorerFilterFields is a list of supported Log Explorer filter fields.
+var LogExplorerFilterFields = []LogExplorerFilterField{
```
anthony-gomez-fastly, line 39: so we are using these downstream?

anthony-gomez-fastly (approved):
LGTM, just one question

fastly/observability_log_explorer.go
anthony-gomez-fastly: ah, i read the description

rcaril (approved):
LGTM!

